PR #34, opened 2026-03-17
Fix generic project representation and edition

Shadow-MMN:
- Corrected code according to review
- Updated non-breaking dependencies


tupui (approved):
Getting this in for now. Thanks! I will test an edit when I get more time but the visual changes are already welcomed.

Issue #18, opened 2026-03-09
Evaluate support for Radicle as another VCS


tupui:
xref https://github.com/tupui/soroban-versioning/issues/343

Septimus4:
Hello @tupui do you want me to start looking into this Issue or is there anything you would prefer I look into first ?

tupui:
Hell yes!! We might need a worker as I am not sure if we can do it all with an API approach. If the worker is needed, all good as this is what we discussed on the other issue.

tupui:
In the end this could look pretty straightforward because there are public seeder/explorers like this:

https://radicle.network/nodes/iris.radicle.xyz/rad:zssaAF91kxuquZmZCV2SiK2FNX6s

There you can clone using httpd. E.g. `git clone https://iris.radicle.xyz/zssaAF91kxuquZmZCV2SiK2FNX6s.git tansu`

PatrickKish1:
@PatrickKish1 has applied to work on this issue as part of the **Stellar Wave Program**'s 5th wave.



> I want to contribute to the issue. This is a product-architecture problem, not just an engineering one, when the source of truth isn't GitHub itself, but rather the repository that contains the code for the entire application, contributors and auditors still require trustworthy project history, and that means defining a model of timeline, stable identifiers and a UI that consistently displays that history. I've written similar “decouple from vendor UI” stuff in the past, where it ended up being a pipeline to maintain instead of just a one off script because it failed to create duplicate sources of truth and had idempotent migrations. I will maintain the implementation consistent with the current storage patterns, and add tests to specific ordering, deduplication and empty state scenarios. I also checked on linked issues #343 and #81


ℹ️ **Repo Maintainers**: To accept this application, [review their application](https://www.drips.network/wave/maintainers/issues/fdf117d5-b316-453f-bcab-f97a6f8b0028) or assign @PatrickKish1 to this issue.

tupui:
Prio is for @Septimus4 otherwise you get it @Oluwatos94 if he is not interested. thanks

nonso7:
@nonso7 has applied to work on this issue as part of the **Stellar Wave Program**'s 5th wave.



> Can i work on this issue


ℹ️ **Repo Maintainers**: To accept this application, [review their application](https://www.drips.network/wave/maintainers/issues/fdf117d5-b316-453f-bcab-f97a6f8b0028) or assign @nonso7 to this issue.

Septimus4:
[...]

PR #41, opened 2026-03-21
Add Cloudflare Git proxy backend and local metadata API fallback

### LINKED ISSUE #22: History without GitHub
xref https://github.com/tupui/soroban-versioning/issues/81

Septimus4:
Add two git metadata backends for the dapp:

1. A local Astro API fallback that uses the `git` binary and supports arbitrary public HTTP(S) repositories.
2. A Cloudflare Worker backend that uses provider APIs for GitHub, GitLab, Bitbucket, Codeberg, and Gitea.

This also refactors the dapp to use repository URLs directly instead of assuming GitHub owner/repo pairs throughout the UI and state layer.

### Local/container-backed git support
- Added a new Astro endpoint at git.ts
- Supports:
  - `history`
  - `commit`
  - `latest-hash`
  - `readme`
- Uses shallow bare clones plus caching to avoid repeated full fetches
- Accepts public `http://` and `https://` repository URLs
- Keeps SSH support restricted to known hosts plus `GIT_ALLOWED_HOSTS`
- Blocks localhost, private IP ranges, and non-standard HTTP(S) ports to avoid private-network proxy behavior

### Cloudflare Worker backend
- Added git-proxy
- Implements the same API contract as the local Astro endpoint
- Supports provider-backed metadata for:
  - GitHub
  - GitLab
  - Bitbucket
  - Codeberg
  - Gitea
- Optional provider credentials can be configured via Cloudflare secrets

### Frontend and state refactor
- Switched project repo state from `{ author, repository }` to `repoUrl`
- Updated commit history, latest commit, readme, and contribution metrics to use repository URLs
- Removed GitHub-only assumptions in project creation/update flows
- Added provider-aware release links where supported
- Updated some GitHub-only copy/icons to be repository-generic

### Validation and configuration
- Repository URL validation now accepts:
  - public `http://` / `https://`
  - `git@host:owner/repo.git`
- Added:
  - `PUBLIC_GIT_PROXY_URL`
  - `GIT_ALLOWED_HOSTS`
- Ignored repo-root .env
- Updated worker config/docs so wrangler.toml is safe to commit by default and CORS overrides are documented

Added test coverage for:
- local Astro git API validation and history behavior
- worker URL parsing and provider response normalization

## Notes

- If `PUBLIC_GIT_PROXY_URL` is set, the dapp uses the Cloudflare Worker backend.
- If `PUBLIC_GIT_PROXY_URL` is unset, the dapp falls back to the local Astro `/api/git` endpoint.
- `CORS_ALLOWED_ORIGINS` in the worker is optional; if omitted, the worker uses its built-in defaults.

I looked at replacing everything with the Cloudflare Worker, but that only really works well for supported providers and their APIs. It doesn’t cover arbitrary git hosts in a general way. For that, I need a containerized setup that can run git directly, which adds more moving parts and more cost. Keeping both paths felt like the better tradeoff: use the Worker where it fits, and keep the container-backed fallback for the broader git case.


Septimus4:
I don't know what is Drips Wave so no :)
No pressure don't worry

tupui:
Thanks for keeping this up to date.

Question about the 2 modes. Is the "local" version purely running client side? If that's the case, why not just having this mode?

Septimus4:
The main reasons are deployment complexity and cost.

The Worker version is easy to configure and deploy, and it is also the cheaper option. At low usage it could even fit within the free tier, and if usage grows the standard paid Workers plan should still be enough for most practical usage.

The local/container approach is heavier: it needs a runtime with git, plus caching, readiness, temp storage, and more security considerations. It is also harder to price because Containers have no free tier, and once the included usage is exhausted the cost depends a lot on the actual runtime behavior and configuration (you could optimised based on type of usage, I won't go into details now, but I can write about this if you want).

Supporting both also gives a nice hybrid path. I would expect most requests in practice to be for the main git providers, so those can go through the Worker/API path, which is cheaper.
If we used only the local/container path, then even those mainstream provider requests would still consume container runtime and we would pay for them there as well. With the mixed approach, the common case stays on the cheaper and easier path, and the container path is only needed for repositories that are not covered by the supported provider APIs.

tupui:
> because Containers have no free tier

Sorry I am not following. Right now the dapp is purely running client side. There is no backend or cloud function. It is deployed on Netlify and thanks to that I can also deploy it on IPFS via https://xlm.sh . The only backend there is now is a cloudflare worker for IPFS handling.

Ideally, I want to limit the use of cloudflare and try to move everything to run client side. For Git, since we do not need a token, I would prefer to do everything on the user side.

Septimus4:
yeah exactly that is the issue, for the local path you need a backend for the git runtime, you can not run that on the client side. So the best way I found is  a cloudflare containers https://developers.cloudflare.com/containers/

tupui:
I think we can do with a small wasm for git no? E.g. I could see this https://github.com/petersalomonsen/wasm-git (not saying to use this one)

tupui:
Like there seem to be smaller things https://github.com/powercord-org/simple-git-wasm

In the end we want the bare minimum.

Septimus4:
it's been some time since I did those research and I remember looking into isomorphic-git but there were some issues for the flow, I would have to look again for the exact problems, maybe CORS related. I did not remember looking into wasm stuff, I can check

tupui:
Thanks. Otherwise, I would just keep the cloudflare worker option since the other option also involves cloudflare but with containers. Support for the most common providers is enough for now.

Septimus4:
Yeah I checked rapidly the isomorphic-git doc highlight CORS issue and the need for a proxy, which defeat the purpose for us.
For [wasm-git](https://github.com/petersalomonsen/wasm-git) I would expect the same issue, and from the doc they list: 
1. Sync version
2. Async versions (with Asyncify support)
3. OPFS versions (with WASMFS and OPFS support)

For  Sync and OPFS versions it's clearly stating the needs for workers, and it's unclear for the Async,  but even if the Async can run fully in the browser you woudl still need a proxy for CORS, at least that is my understanding.

Septimus4:
@tupui I cleaned-up the code, I also updated the flow to add or edit a project, to better match and represent the multi provider support (headings, placeholder, icons etc), I also changes the flow a bit to ask for the url on the first modal and adapt the flow based on the provider detected.

I still need to do some testing, I have not tested all providers, I will finish testing at least one project by providers. The code should be okay to review.

I kept the commits with the local git and container path, in case it could be useful later. I can squash them if you prefer

tupui:
Great thanks for the update! @Shadow-MMN if you have some time to have a look too that would be helpful.

Shadow-MMN:
> Great thanks for the update! @Shadow-MMN if you have some time to have a look too that would be helpful.

Kinda busy rn, I will take a look at it when I am done with what I am doing. Expect a review this night.

dapp/src/utils/editLinkFunctions.ts
Shadow-MMN, line 1: `parseRepositoryUrl` validates host via hostname but ignores port. URLs like `https://github.com:8443/owner/repo` pass host validation but would attempt a non‑standard port connection. Can we check `parsedUrl.port` is empty or 443 then reject others after parsing?.

Shadow-MMN, line 1: In `isSupportedRepositoryUrl()`, the validation message suggests broader SSH support than what is actually implemented. Right now only SCP-style SSH (`git@host:repo`) is supported, while `ssh://` URLs are not.

To keep behavior consistent and avoid misleading expectations, it would be better to either adjust the message/tests to match the current supported formats, or explicitly clarify supported SSH formats in the validation logic.

This is mainly a consistency issue between implementation and documentation rather than a functional bug.

dapp/src/service/RepositoryMetadataService.ts
Shadow-MMN, line 1: Provider API calls in RepositoryMetadataService appear to execute without retry, caching, or circuit-breaker logic. This means transient network failures or rate limits will directly surface to the UI. While not blocking, adding basic retry/backoff or short-lived caching could improve reliability.

Suggested fix :  Re‑introduce a fallback strategy (e.g., retry with backoff, optional worker proxy via `PUBLIC_GIT_PROXY_URL`, or in‑memory cache of recent responses).

dapp/src/service/GithubService.ts
Shadow-MMN, line 1: I noticed that this file appears to contain the same logic as `RepositoryMetadataService.ts`. Is there a specific reason for maintaining both implementations separately?
If there isn’t a deliberate distinction between them, it might be better to consolidate them into a single source of truth and update all imports accordingly. This would help reduce duplication and make future maintenance easier.

dapp/src/utils/editLinkFunctions.ts
```diff
@@ -13,43 +13,254 @@ export function convertGitHubLink(link: string | null | undefined): string {
...
+  }
+
+  if (parsed.host === "codeberg.org" || parsed.host === "gitea.com") {
+    return `${parsed.normalizedUrl}/releases`;
   }
+
   return undefined;
 }
+
+export function getAuthorRepo(repoUrl: string | null | undefined): {
```
Shadow-MMN: This functions is longer in use after refactor to `repoUrl`. I suggest we delete it.

dapp/src/service/RepositoryMetadataService.ts
Shadow-MMN, line 1: Owner/repo segments are currently interpolated directly into provider API URLs without being URL-encoded. This can lead to malformed requests when special characters (such as spaces or Unicode characters) are present, and may also cause unexpected behavior if a provider interprets the path incorrectly.

It would be safer and more consistent to encode each path segment using `encodeURIComponent`, similar to how it’s already handled in the GitLab and Bitbucket readme logic.

dapp/src/utils/editLinkFunctions.test.ts
Shadow-MMN, line 1: Tests only cover basic positive cases. We can add comprehensive unit tests for parseRepositoryUrl and isSupportedRepositoryUrl covering valid/invalid HTTPS, SSH, ports, IP ranges, and special characters.

Shadow-MMN (changes requested):
In the `dapp/src/utils/utils.ts` When ORG_GITHUB is set, `extractConfigData` always builds `https://github.com/${ORG_GITHUBregardless}` of actual provider. Non‑GitHub repos get a broken "GitHub" link.
I think it would be better if we derive the `officals.githublink` from `project.config.url` (the canonical repo URL). If `ORG_GITHUB` must be preserved, construct the URL using the provider derived from `project.config.url` instead of hardcoding `github.com`.

Apart from all these that I pointed out you did a very wonderful job. I love the way you used the switch-statement for the provider, how you isolated the repository metadata fetching in service layer, how the `formatCommits` normalizes provider responses into `FormattedCommit` used throughout UI.

The app loads well and nothing breaks a far as I have tested. Happy coding 😊.

dapp/src/utils/editLinkFunctions.test.ts
Septimus4: Added and extended some tests

dapp/src/service/RepositoryMetadataService.ts
Septimus4: Added getEncodedRepositorySegments

dapp/src/utils/editLinkFunctions.ts
Septimus4: Done

dapp/src/service/GithubService.ts
Septimus4: It was not even used, it just stayed from the refactor, my bad

dapp/src/utils/editLinkFunctions.ts
Septimus4: Left over from the full local git support, fixed my making clear we support git@host:repo not ssh://

dapp/src/service/RepositoryMetadataService.ts
Septimus4: retry + short-lived in-memory should be enough for availability purposes.
But if ti was intended to help with rate limits, CORS then thew full worker proxy is better. @Shadow-MMN tell me what you had in mind

dapp/src/utils/editLinkFunctions.ts
Shadow-MMN: Thanks for the clarification. That makes sense.

dapp/src/service/RepositoryMetadataService.ts
Shadow-MMN: Just to clarify, I was referring to basic client-side resilience, not replacing the worker proxy.

Mainly:

- a small retry (1–2 attempts with backoff)
- short-lived in-memory caching to avoid duplicate calls

This is just to handle transient issues like network glitches, not CORS or rate limiting at scale.
I agree the Worker proxy is the right place for provider-level concerns.

So basically:
- Worker → provider/CORS abstraction
- Client → lightweight retry + caching for stability

dapp/src/utils/editLinkFunctions.ts
```diff
@@ -13,43 +13,321 @@ export function convertGitHubLink(link: string | null | undefined): string {
...
+  projectPath: string;
+  repoName: string;
+  owner: string;
+}
+
+function hasValidRepositoryPathSegments(
+  host: string,
+  segments: string[],
+): boolean {
+  if (host === "gitlab.com") {
```
Shadow-MMN, line 105: This accepts any GitLab URL with at least two path segments, so URLs like `https://gitlab.com/group/project/tree/main` or `.../blob/main/file` are treated as valid repository roots. That will make the new metadata fetches hit the wrong project path. This needs a tighter parser plus tests for GitLab tree/blob links.

Shadow-MMN (changes requested):
The refactor is in the right direction and most of the review feedback looks addressed. We still need one more pass after the URL parser is tightened. I ran the focused vitest files locally for the new helper/service tests and they passed , great job. One more change and we are good to go. Happy Coding 😊.

dapp/src/utils/editLinkFunctions.test.ts
Shadow-MMN: Great job 💪🏾

dapp/src/service/RepositoryMetadataService.ts
Shadow-MMN: Nice work 👍🏾

Shadow-MMN (approved):
The parser now normalizes GitLab tree/blob URLs to the repository root, and the tests cover it. Every other thing seems fine to me and I am happy to approve.
@tupui , I am approving this PR as the issues I identified have been addressed. Please proceed with your review when convenient.

Shadow-MMN:
Wonderful job @Septimus4 , you are a very awesome dev.

tupui:
Ok so if you're both happy now I can give this a try :)

tupui (commented):
Just where is the worker code I have to push on Cloudflare?

Ideally we would put this in the worker folder with the other one. We could consider using the same worker for both IPFS and Git. A bit neutral here.

Septimus4:
Since we chose to go the API way there was no need for worker logic anymore, everything is client side.
The only code I could add to the Worker is a proxy to avoid CORS issues, like a fall back if the client gets a CORS related error.
It can be done in this PR or the next, or even wait to see if it's necessary

tupui:
Oh great news. Somehow I still thought it was still needed to have a worker to do the git cloning ops from our discussions.

tupui:
This https://github.com/Consulting-Manao/tansu/pull/41#issuecomment-4242584957

Maybe we did not understand each other.

tupui:
@Septimus4 what is the status here? Cloudflare, no cloudflare? I am still not sure what we are doing then here based on what we last said.

Septimus4:
Cloudfare container was onpy needed for git runtime, since we only integrate with APIs nor cloudfare container or worker are useful, client side should be enough in most case.

The potential need for Cloudfare worker would be handling CORS issues for more restricted providers for now Github, bitbucket are permissive. For Gitlab, gitea instances it depends on each providers.

We should look into merge this PR, it's already a lot of changes.

My next work could be to add to the IPFS Cloudfare worker a proxy to bypass CORS errors when a client encounter the issue.

tupui:
Ok so if we are saying that all the changes are making things more generic in terms of provider now and that the git logic is contained in `dapp/src/service/RepositoryMetadataService.ts` then I am fine 👍 

One of my long term dream is still to support Radicle #18 or other very different systems. That might mean a worker on Cloudflare at that point.

All that to say that if merging this is also some prep work towards that generic goal, then let's go. I just want to avoid us being locked into yet another system. That's the primary reason for this in the first place, to let people use what they want. So that's also meaning we should not care much about the provider too.

Septimus4:
Yes even if we move to always use a Worker to query the git API later this PR helps because it gives the base to support multiples P[roviders, and formatCommits gives common interface to give the FE. I don't know how Radicle looks exactly so not 100% sure, but my guess is as long has you can make a valid formatCommits objects you would not need to change anything FE side. And of course is there is some specificities at least you a a single interface to expend/modify to ensure compability

tupui:
Ok then perfect let's do this 😃 Thank you so much for keeping this up really! Just the conflicts to solve on your side then.

@Shadow-MMN please have another pass if you have time and then I am happy to get this in. It's a big one so I just want to be sure and not rush it (I know we are past rushing, I am used to SciPy's rushing 😉 I had some PR last a year with active convos lol)

Shadow-MMN:
@Septimus4 Excellent job so far.
The `githubRepoUrl` is currently persisted exactly as entered, so GitLab `/-/blob/` or `/-/tree/` URLs can be stored instead of the repository root URL. This later breaks consumers like `ProjectReadme.astro` and `ReadMoreModal.tsx` that expect a cloneable repository URL. Can we consider normalizing the URL before persisting it.

Septimus4:
@Shadow-MMN Fixed

Shadow-MMN:
> @Shadow-MMN Fixed

Ok , let me test the changes and get back to you

dapp/src/utils/utils.ts
```diff
@@ -28,6 +33,14 @@ export function truncateMiddle(str: string, maxLength: number): string {
 export function extractConfigData(tomlData: any, project: Project) {
   const fullName = tomlData.DOCUMENTATION?.ORG_DBA || project.name;
   const projectType = tomlData.PROJECT_TYPE || "SOFTWARE";
+  const canonicalRepositoryUrl = isSupportedRepositoryUrl(project.config.url)
+    ? normalizeRepositoryUrl(project.config.url) || ""
```
Shadow-MMN: We are almost there, great job so far. We are getting there.
`normalizeRepoUrl` now drops any `project.config.url` that is not a supported provider URL before fallback. This causes legacy or unsupported repository URLs to lose their repo link/icon in the dashboard and modal, whereas the previous behavior still displayed the raw URL. Can we consider preserving the original `project.config.url` when normalization fails.

tupui:
@Shadow-MMN @Septimus4 feel free to ping me once this is ready to merge. I will only quickly go over just in case but trust you on this one mostly 😃

Shadow-MMN (approved):
Image

Screenshot 2026-05-06 at 20 38 41


Works well. I was able to test this successfully with GitLab. Great job, and thank you for contributing to Tansu.
Happy coding 😊
@tupui, could you also take a look on your side? Everything looks good to me.

tupui:
Thanks a lot both. I will just go ahead and merge now. Will test tomorrow on my side. I trust you both by now ;)

tupui:
@Septimus4 you might have to open a ticket with Drips to get points. Not totally sure there I know they also check a bit manually things.

Septimus4:
Thanks for the reviews and guidance.
Happy to see this feature added.

dapp/src/components/page/project/ProjectInfo.astro
```diff
@@ -587,6 +587,85 @@ import Button from "components/utils/Button";
           }
         }
 
+        if (syncStatusContainer && isSoftwareProject) {
```
tupui, line 590: @Shadow-MMN please be more careful next time when reviewing. This was deleted in another PR and not needed here as it's duplicating a feature. I am removing this now myself.

Septimus4: Sorry probably comes from the old branch or a merge I did

tupui: No worries, this is why we do code reviews. It was also a big one so all good. What matters really is that we did not break anything 😅

Shadow-MMN: > @Shadow-MMN please be more careful next time when reviewing. This was deleted in another PR and not needed here as it's duplicating a feature. I am removing this now myself.

Ok I will

PR #42, opened 2026-03-24
fix(wallet): unify state management and resolve race conditions during init

### LINKED ISSUE #38: Wallet connection state
When we are connected, sometimes we cannot do actions such as creating a new project as we get asked to connect a wallet again. This is a state management issue and should not happen. It should not be confusing. We are either connected or we are not.

KaruG1999:
## Problem
On refresh, the UI would show "Connect Wallet" to users with an active session, or open the Create Project modal without a valid public key. 

Three layers held wallet state independently (module singleton, Nanostore atom, and localStorage) with diverging initial values — `undefined`, `""`, and a persisted string. Any component reading one layer got a different answer depending on the render cycle.

## What changed

**store.ts / walletService.ts**
- **`connectedPublicKey` initial value:** `""` → `undefined`. Necessary so `!atom.get()` reliably means "not connected" at every phase, including before init completes.
- **Added `walletInitialized` atom:** Set to `true` only after the extension handshake settles. This gates the Connect button and the modal to distinguish "disconnected" from "still checking".
- **`initializeConnection()` is now async:** Sets localStorage key optimistically, then verifies with `StellarWalletsKit.getAddress()`. If the extension revokes or returns a different address, state corrects automatically.

**ConnectWallet.astro**
- Button disabled during handshake, enabled on `walletInitialized`.
- Correction handled in the subscription: if the confirmed key differs from what's displayed, it updates the button or dispatches `walletDisconnected`.

**CreateProjectModal.tsx**
- Replaced `useState([loadedPublicKey() || ""])` — which captured a stale value at mount. Now seeds the maintainer field via `useEffect` once `connectedPublicKey` resolves.
- Step 1 "Next" now checks wallet connection. Previously the form let users reach step 4 before failing.

**NavbarSearch.tsx**
- Replaced `loadedPublicKey()` + window event listeners with `useStore(connectedPublicKey)`. The atom subscription is the reactive equivalent, removing manual event wiring.

**ProjectList.jsx (Path B)**
- The `sessionStorage` trigger had no wallet guard. Fixed with a one-shot `walletInitialized` subscriber so the modal only opens after init.
- `CreateProjectModal` mount is now gated by `isWalletReady` to prevent mounting with unresolved state.

**ProjectCard.tsx / ReadMoreModal.tsx**
- `logoImageLink !== undefined` passed for `""` (set by `extractConfigData` as fallback). `convertGitHubLink("")` returns `""` → ``. Changed to a truthiness check.

## What was not changed
Service files (`ContractService`, `FlowService`, etc.) read `loadedPublicKey()` at transaction time, not at render time — no stale closure problem there. `loadedPublicKey` remains exported for those callers to avoid unnecessary refactors in stable business logic.

---

## Visual Evidence

tansu1
tansu2
tansu3

---

## Test Plan
- [ ] **Persisted Session:** Refresh page with wallet connected → button shows Profile, no flash.
- [ ] **Clean State:** Refresh page with wallet disconnected → Connect button shows, no stale state.
- [ ] **Early Guard:** Open "Add Project" disconnected → toast error appears at step 1.
- [ ] **Reactivity:** Open "Add Project" connected → maintainer field auto-populates correctly as soon as the handshake resolves.
- [ ] **Console Health:** Browser console shows no `` warnings and no hydration errors.
---

###  Revisions
The solution maintains the existing project architecture (Astro + Nanostores) without adding external dependencies
**I am fully available for any feedback, technical adjustments, or further refinements to ensure this aligns with the team's standards.** Thank youuu

closes #38


tupui (commented):
Thanks a lot for the work! @Shadow-MMN please have a look.

Shadow-MMN (commented):
LGTM, nice work honestly. You are awesome.

tupui (approved):
Getting this in now. There will be some tests to fix still. Can be a follow up.

PR #44, opened 2026-03-26
fix: Cannot create a project after creating an account

### LINKED ISSUE #35: Cannot create a project after creating an account
After creating an account, we must reload the page otherwise we get a wrong error message saying we are not connected. We need to reconnect and then it works.

Oluwatos94:
## Summary:

1. NavbarSearch.tsx: Replaced the event-based isWalletConnected state (which relied on
  walletConnected/walletDisconnected custom events and could have timing issues) with a direct reactive subscription to
  the connectedPublicKey nanostore via useStore. This means isWalletConnected is now always in sync with the actual
  wallet connection state as soon as setConnection() is called during joining, the "Add Project" button becomes
  enabled.

  2. CreateProjectModal.tsx: Removed the redundant dynamic import("@service/walletService") inside
  handleRegisterProject. The static import of loadedPublicKey at the top of the file is sufficient  the dynamic import
  was shadowing it unnecessarily and could potentially resolve to a different module instance in some bundler
  configurations.

  Root cause: When the user creates an account (joins) via the JoinCommunityButton flow without a prior wallet
  connection, setConnection() is called and updates the nanostores, but the event-based isWalletConnected in
  NavbarSearch could be stale if there was a timing issue with the walletConnected event listener. Using useStore
  eliminates this race condition entirely.

closes #35

### DIFF
--- a/dapp/src/components/NavbarSearch.tsx
+++ b/dapp/src/components/NavbarSearch.tsx
@@ -1,7 +1,8 @@
 import { useState, useEffect } from "react";
+import { useStore } from "@nanostores/react";
 import Button from "./utils/Button";
 
-import { loadedPublicKey } from "../service/walletService";
+import { connectedPublicKey } from "../utils/store";
 import { toast } from "../utils/utils";
 
 // Constants for URL handling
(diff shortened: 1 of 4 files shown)


tupui (approved):
Thanks 🙏

PR #48, opened 2026-03-27
feat: update social cards to use GitHub README banner image

### LINKED ISSUE #33: Update social card
The social cards of the website and dApp are not looking nice. We should use the same card we have now on GitHub and the Readme.

nonso7:
Closes #33

---

Completed

You can reach out to me if not satisfied


tupui (commented):
Thanks 🙏 do we still need the other png?

Also not sure we need to duplicate the ressource as in the end we use a full URL no? Would be still ok like this.

tupui (commented):
I left a comment to clarify hopefully.

The other point is that you added the banner png to both site. Is this necessary? Or could we just have that on one site and point to the other one? Maybe cors issues.

tupui (approved):
Let's just get this in now. Thanks

PR #49, opened 2026-03-28
Dual pinning on IPFS

### LINKED ISSUE #39: Dual pinning on IPFS
Right now we use Storacha for storing data on IPFS.

Storacha has had a few issues recently making it unreliable. Hence we need to do a dual pinning to backup the data.

We just need to add this on the saving part. Multiple paths to modify and should be of course done in a centralized manner. Project creation/update, membership, proposal, etc.

On the reading part we should already have some fallback logic to use https://filebase.com/, let's use them as well for the backup. If there is a need for an API key to upload to the space, ping me. We will want to use a similar strategy as with Storacha if really needed. But hopefully they would have something simpler we could use? To check.

Richiey1:
Closes #39

## Summary
- Adds Filebase as backup IPFS pin provider alongside Storacha
- Dual pinning is fire-and-forget (doesn't block user flows)
- Graceful degradation if FILEBASE_API_KEY not configured

## Changes
- `dapp/src/utils/dualPin.ts` — New utility with `pinToFilebase`, `dualPin`, `dualPinVerified`
- `dapp/src/service/FlowService.ts` — Integrated Filebase backup pin after Storacha upload
- `dapp/.env.example` — Added `FILEBASE_API_KEY` variable

## How it works
1. Primary upload goes to Storacha (existing flow, unchanged)
2. After Storacha succeeds, CID is pinned to Filebase in the background
3. If Filebase fails or key is not set, the upload still succeeds (Storacha is primary)
4. On reading side, existing fallback logic (filebase → storacha gateways) already works

## Configuration
Set `FILEBASE_API_KEY` in `.env` to enable dual pinning. If not set, uploads still work via Storacha alone.

### DIFF
--- a/dapp/.env.example
+++ b/dapp/.env.example
@@ -10,3 +10,6 @@ PUBLIC_TANSU_OWNER_ID="GCILP4HWE2QGEO4KUMOZ6S6J3A46W47EVCGZW2YPYCPH5CQF6EACNBCN"
 
 # IPFS upload flows
 PUBLIC_DELEGATION_API_URL="https://ipfs-testnet.tansu.dev"
+
+# Dual IPFS pinning — Filebase backup (optional but recommended for redundancy)
+FILEBASE_API_KEY=""
--- a/dapp/src/service/FlowService.ts
+++ b/dapp/src/service/FlowService.ts
@@ -2,6 +2,7 @@ import { calculateDirectoryCid } from "../utils/ipfsFunctions";
 import { create } from "@storacha/client";
 import * as Delegation from "@ucanto/core/delegation";
 import type { OutcomeContract } from "../types/proposal";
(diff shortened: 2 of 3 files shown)


Richiey1:
Hey @tupui, I've added dual IPFS pinning with Filebase as backup alongside Storacha. The pinning is fire-and-forget so it doesn't slow down the upload flow. Let me know if you'd like any changes or if the Filebase API key needs to be set up!

tupui (commented):
Thanks for working on this 🙏 I left a comment.

tupui:
@Richiey1 friendly reminder

Shadow-MMN (commented):
Nice code bro, but it is missing a few things like retry handling if filebase is not currently up . Apart from that everything works well and solves the issue linked to it.
LGTM

PR #56, opened 2026-04-06
feat: migrate IPFS to dual-provider architecture (Filebase & Pinata)

Shadow-MMN:
### Description

  This PR migrates the IPFS write path away from Storacha and onto a Cloudflare worker-backed flow using Filebase as the active upload provider, while keeping Pinata support in place
  behind a feature flag for future paid-plan enablement.

  The goal is to make the proposal upload path stable for launch now, while preserving a clean path to enable Pinata backup/pinning later without another architectural rewrite.

  ### Key Changes

  - Replaced the previous Storacha delegation flow with a Cloudflare IPFS proxy worker in dapp/workers/ipfs-delegation.
  - Removed Storacha / UCAN upload dependencies from the frontend and worker flow.
  - Kept deterministic CID generation by packaging uploads into a CAR locally before sending them to the worker.
  - Reworked all upload flows to:
      - calculate the CID locally
      - sign the on-chain transaction once
      - upload the CAR through the worker
      - verify the returned CID before submitting the signed transaction
  - Updated the worker to:
      - verify the signed transaction
      - verify the CAR root CID matches the expected CID
      - upload to Filebase as the active blocking path
      - optionally pin the CID on Pinata in the background when ENABLE_PINATA_PINNING=true
  - Updated IPFS reads to use:
      - Filebase first
      - https://ipfs.io/ipfs/... as fallback
  - Added retry handling and clearer upload error propagation from the worker to the UI.
  - Cleaned review noise and local tooling issues:
      - restored explicit local Wrangler port usage
      - ignored .wrangler temp files in Prettier

  ### Current Behavior

  - Filebase is the active upload provider.
  - Pinata support remains implemented but disabled by default behind ENABLE_PINATA_PINNING.
  - This means the current launch path works with Filebase only, while future paid Pinata enablement can reuse the existing CAR-based path.

  ### Verification

  - cd dapp && bun run lint
  - cd dapp/workers/ipfs-delegation && bunx tsc --noEmit
  - local Wrangler worker test via bun run test
  - manual local dapp verification with the worker running on http://localhost:8787

  ### Deployment Notes

  Set these Cloudflare Worker secrets/config:

  1. FILEBASE_TOKEN
  2. ENABLE_PINATA_PINNING
  3. PINATA_JWT if Pinata pinning is enabled
  4. PINATA_GROUP_ID if Pinata grouping is needed

  The dapp environment should continue using:
  PUBLIC_DELEGATION_API_URL="https://your-worker-url"


dapp/workers/ipfs-delegation/src/index.ts
```diff
@@ -1,217 +1,231 @@
...
+          name: "filebase",
+          cid: returnedCid,
+        };
+      } catch (e: any) {
+        return { ok: false, error: e.message, name: "filebase" };
       }
+    };
+
+    // --- 2. Backup: Pinata CAR Upload ---
+    const pinataUpload = async (retries = 2): Promise<any> => {
```
tupui: To be clear, both upload must happen at the same time, not one after the other. Otherwise it takes too much time unnecessarily

dapp/workers/ipfs-delegation/README.md
tupui: Why removing??

dapp/src/service/FlowService.ts
```diff
@@ -54,74 +46,68 @@ interface CreateProjectFlowParams {
 }
 
 /**
- * Upload files to IPFS using delegation
+ * Upload files to IPFS using the Proxy Worker.
+ * Accepts pre-calculated CID and CAR blob to ensure absolute consistency.
  */
-export async function uploadWithDelegation({
```
tupui: As said elsewhere, don't change the API here. It was correct before. We want to have a message signature and check it in the worker.

tupui (changes requested):

tupui (changes requested):
Left some comments

dapp/workers/ipfs-delegation/src/index.ts
```diff
@@ -53,165 +57,311 @@ function getCorsHeaders(origin: string | null): Record<string, string> {
...
-    const delegation = await client.createDelegation(
-      audience,
-      abilities as any,
-      {
-        expiration,
-      },
-    );
+  let lastCid: string | undefined;
+  for await (const block of reader.blocks()) {
+    lastCid = block.cid.toString();
```
Oluwatos94: The reader.blocks() iterates in internal encoding order, not by role. If roots[0] is undefined, the fallback assigns the CID of whatever block happens to be last this is not guaranteed to be the directory root and could be any leaf node. 

I would suggest If roots[0] is missing, throw immediately don't iterate:                                                         
const roots = await reader.getRoots();
  if (!roots[0]) throw new Error("CAR file has no declared root");                                                        
return roots[0].toString();

dapp/workers/ipfs-delegation/src/index.ts
```diff
@@ -53,165 +57,311 @@ function getCorsHeaders(origin: string | null): Record<string, string> {
...
+          },
+          body: formData,
+        });
+
+        if (!res.ok) {
+          throw new Error(`Filebase HTTP ${res.status}`);
+        }
+
+        const text = await res.text();
+        if (text && !text.includes(body.cid)) {
```
Oluwatos94: Substring search on raw response text is not adequate, a CID appearing in an error message field would pass validation. This should parse the JSON and check the correct field.

dapp/src/utils/ipfsFunctions.ts
```diff
@@ -230,30 +225,150 @@ export const getProposalLinkFromIpfs = (cid: string): string =>
...
+    throw new Error("Missing signed transaction for IPFS upload");
+  }
+
+  if (!(carBlob instanceof Blob) || carBlob.size === 0) {
+    throw new Error("Invalid CAR blob for IPFS upload");
+  }
+
+  let binary = "";
+  for (const byte of new Uint8Array(await carBlob.arrayBuffer())) {
+    binary += String.fromCharCode(byte);
```
Oluwatos94: Each += allocates a new string. This will probably freeze the browser for files over 1MB. I feel you can replace with chunked encoding or maybe send as raw binary. I may prefer to handle it this way.

const bytes = new Uint8Array(await carBlob.arrayBuffer());
  let binary = "";
  for (let i = 0; i < bytes.length; i += 8192) {
    binary += String.fromCharCode(...bytes.subarray(i, i + 8192));
  }  
const car = btoa(binary);

Oluwatos94:
@Shadow-MMN, Well done bro. You did a nice job. Checking your code I was able to spot on logics I feel might need improvement... 

And lastly, I notice validateSignedTransaction and calculateCidFromCar are the two security gates of the worker (signature verification and CID integrity check). Neither has a unit test. packFilesToCar on the client side also has no test. At minimum these three functions need coverage under bun test. 

Well done once again. ))

Shadow-MMN:
> The reader.blocks() iterates in internal encoding order, not by role. If roots[0] is undefined, the fallback assigns the CID of whatever block happens to be last this is not guaranteed to be the directory root and could be any leaf node.
> 
> I would suggest If roots[0] is missing, throw immediately don't iterate: const roots = await reader.getRoots(); if (!roots[0]) throw new Error("CAR file has no declared root"); return roots[0].toString();

Thank you for the review, checking them out now

Shadow-MMN:
> @Shadow-MMN, Well done bro. You did a nice job. Checking your code I was able to spot on logics I feel might need improvement...
> 
> And lastly, I notice validateSignedTransaction and calculateCidFromCar are the two security gates of the worker (signature verification and CID integrity check). Neither has a unit test. packFilesToCar on the client side also has no test. At minimum these three functions need coverage under bun test.
> 
> Well done once again. ))


I have implemented the suggested changes, can you review it and tell you what you think. Thank again for taking your time to review this pr. ☺️

tupui (approved):

tupui:
Thank you both for the work 🚀

PR #80, opened 2026-04-17
fix(dapp): resolve netlify build failure and clean type checks

Shadow-MMN:
## Summary

This PR fixes the dapp build failure seen on Netlify and clears the `bun run check` / `bun run lint` issues without changing current app behavior.

## What changed

- normalized large inline Astro client scripts to avoid TS-only syntax in CI-sensitive bundling paths
- fixed strict type issues in Playwright config, chart tooltip formatting, wallet/public-key state handling, and React error boundary overrides
- tightened typing for terms summary parsing and anonymous voting helpers
- fixed CAR blob typing in both the dapp IPFS helpers and the IPFS delegation worker
- removed unused code that was contributing to lint/check noise

## Why

Netlify was failing during bundling around `ProjectInfo.astro`, while local runtime behavior remained fine. The fix keeps the current UI/flow intact and focuses on compatibility with the production build pipeline plus stricter static checks.

## Validation

- `cd dapp && bun run check`
- `cd dapp && bun run build`
- `cd dapp && bun run lint`

All passed locally.

## Notes

- changes were kept behavior-preserving and limited to build/check compatibility
- left unrelated existing worktree changes untouched

### DIFF
--- a/dapp/playwright.config.ts
+++ b/dapp/playwright.config.ts
@@ -14,14 +14,12 @@ export default defineConfig({
   // Performance optimizations
   timeout: 30000,
   expect: { timeout: 5000 },
-  actionTimeout: 5000,
-  navigationTimeout: 10000,
 
   // Fast execution settings
   fullyParallel: true,
   retries: 0,
   // Use more workers locally for speed; keep single worker on CI for stability
-  workers: process.env.CI ? 1 : undefined,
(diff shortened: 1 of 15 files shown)


tupui:
I fixed this now. Closing. Thanks for trying!

Shadow-MMN:
> I fixed this now. Closing. Thanks for trying!

my pleasure

PR #89, opened 2026-04-23
feat: display project_key and project_name on project page

### LINKED ISSUE #67: Project page should show project_key and project_name
On the project page we should see somewhere the project name and project Key which we have on-chain. This would be very helpful.

legend4tech:
### What changed and why

The project page was missing the project name and key from the Soroban contract.
This PR adds them to the page.

### What was added

A new **On-chain Information** panel now appears between the project
header and the Maintainers section. It shows two fields:

- **Project Name** — the raw name string exactly as it was registered in
  the Soroban contract (not the display name override from the TOML file)
- **Project Key** — the full 64-character hex representation of the
  keccak256 hash of the project name, which is the actual storage key
  used on-chain

The Project Key also has a copy-to-clipboard button. Click it and the
icon briefly switches to a checkmark to confirm the copy, then resets.

### Files changed

**`dapp/src/components/page/project/ProjectInfo.astro`**

- Added the `#project-chain-info` panel in the HTML template with two
  labelled fields (`#chain-project-name` and `#chain-project-key`) and a
  clipboard button using the existing `clipboard.svg` icon
- Imported the `projectState` nanostore from `utils/store` to read the
  serialized project key hex
- Wired up the copy button with a click handler that swaps the icon on
  success
- Extended `updateProjectInfo()` to populate both fields — the name comes
  from `loadProjectInfo().name` and the key is parsed from the
  `project_id` field in the `projectState` nanostore. No extra network
  request is needed; both values are already in memory by the time the
  rest of the page loads

### ScreenShot
Screenshot from 2026-04-22 23-44-06

Closes #67

### DIFF
--- a/dapp/src/components/page/project/ProjectInfo.astro
+++ b/dapp/src/components/page/project/ProjectInfo.astro
@@ -106,6 +106,34 @@ import Button from "components/utils/Button";
       </div>
     </div>
     <div
+      id="project-chain-info"
+      class="p-[15px_16px] lg:p-[30px_72px] bg-white flex flex-col gap-5"
+    >
+      <p class="leading-4 text-base text-secondary">On-chain Information</p>
+      <div class="flex flex-col gap-4">
+        <div class="flex flex-col gap-1">
+          <p class="text-xs text-secondary uppercase tracking-wide">Project Name</p>
+          <p id="chain-project-name" class="text-base text-primary font-mono"></p>
+        </div>
+        <div class="flex flex-col gap-1">
+          <p class="text-xs text-secondary uppercase tracking-wide">Project Key</p>
+          <div class="flex items-center gap-2">
+            <code
+              id="chain-project-key"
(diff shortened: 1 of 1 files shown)


tupui (approved):
Thank you @legend4tech I actually like that. I was not sure about the positioning of the information but this solution works. At least for now it provides the value that I want. I might re-order that in the future but for now this is ok. We could also have hidden that a bit behind some other buttons or in tooltips. I will think about that (I might like that). Let's get this in for now.

tupui:
@legend4tech linter is actually failing. Please fix it first.

legend4tech:
> @legend4tech linter is actually failing. Please fix it first.

@tupui  Linting Failure Resolved

tupui:
Thanks again!

legend4tech:
> Thanks again!

You're highly welcome thanks for the Opportunity i really Appreciate

PR #94, opened 2026-04-23
Add conflict of interest list for proposals

### LINKED ISSUE #71: Conflict of interest list
We should have a way to declare conflicts of interests so that specific addresses should not be able to vote at all on a proposal.

People themselves should be able to add their own address and maintainers of the space should be also able to edit the list.

For now I think people making proposals should not be able to add addresses to the list as they could do something bad otherwise like this. We would need some collateral or else at minima to avoid issues as possible.

This change must also be done on the UI. On the proposal page, we need a new button. Maintainers would be able to add a vector of addresses, while anyone else could only add themselves. Pre-fill with the current address logged-in in that case.

sasasamaes:
## Summary

Resolves #71 by allowing addresses to be declared as having a conflict of interest on a proposal so they cannot cast a vote.

- Anyone can add or remove their own address on any proposal
- Maintainers (other than the proposer) can manage the full list
- The proposer is forbidden from modifying the list to prevent abusive exclusion of voters
- Voters on the list cannot call `vote` (panics with `VoterConflicted`)

## Contract changes (`contracts/tansu/src/`)

- New `ProjectKey::ConflictOfInterest(Bytes, u32)` storage key holding a `Vec<Address>` per proposal
- New `DaoTrait` methods: `add_conflict_of_interest`, `remove_conflict_of_interest`, `get_conflict_of_interest`
- `vote()` now blocks voters on the list
- New errors `VoterConflicted (211)`, `ConflictProposerForbidden (212)`, `ConflictOnlySelf (213)` and event `ConflictOfInterestUpdated`
- Regenerated TypeScript bindings in `dapp/packages/tansu`

## UI changes (`dapp/src/`)

- New `ConflictOfInterestModal` with role-aware behavior (maintainer / self-only / proposer read-only)
- New "Conflict of Interest" button on the proposal page for active proposals
- New service helpers: `getConflictOfInterest`, `addConflictOfInterest`, `removeConflictOfInterest`
- Extended contract error messages to cover the three new codes

## Test plan

- [x] `cargo test -p tansu` (58 tests, including the new `conflict_of_interest_flow`)
- [x] `cargo clippy -p tansu --all-targets -- -Dwarnings`
- [x] `cargo fmt -p tansu -- --check`
- [x] `bun run check` (Astro + TS)
- [x] `bun run lint` (prettier, eslint, ts-prune, error mapping validation)
- [x] `bun run test:unit` (vitest – 20 tests)
- [ ] Manual UI smoke test once deployed to testnet


contracts/tansu/src/contract_dao.rs
```diff
@@ -878,17 +1028,36 @@ fn get_voters(env: &Env, project_key: &Bytes, proposal_id: u32) -> Vec<Address>
...
-        ) {
+        if let Some(vote_) = env
+            .storage()
+            .persistent()
+            .get::<types::ProjectKey, types::Vote>(&types::ProjectKey::Vote(
+                project_key.clone(),
+                proposal_id,
+                voter,
+            ))
+        {
```
tupui, line 1050: Please no unnecessary changes anywhere.

contracts/tansu/src/contract_dao.rs
```diff
@@ -869,6 +875,150 @@ impl DaoTrait for Tansu {
...
+        }
+        if proposal.proposer == caller {
+            panic_with_error!(&env, &errors::ContractErrors::ConflictProposerForbidden);
+        }
+
+        let project = get_project(&env, &project_key);
+        let is_maintainer = project.maintainers.contains(&caller);
+        if !is_maintainer && (addresses.len() != 1 || addresses.get_unchecked(0) != caller) {
+            panic_with_error!(&env, &errors::ContractErrors::ConflictOnlySelf);
+        }
```
tupui: Should just use `crate::auth_maintainers(&env, &maintainer, &project_key);`

A maintainer of the project can edit the list.

contracts/tansu/src/contract_dao.rs
```diff
@@ -869,6 +875,150 @@ impl DaoTrait for Tansu {
...
+    ///
+    /// # Arguments
+    /// * `env` - The environment object
+    /// * `project_key` - The project key identifier
+    /// * `proposal_id` - The ID of the proposal
+    ///
+    /// # Returns
+    /// * `Vec<Address>` - Addresses barred from voting on the proposal
+    fn get_conflict_of_interest(env: Env, project_key: Bytes, proposal_id: u32) -> Vec<Address> {
+        get_conflicts(&env, &project_key, proposal_id)
```
tupui: Why having the implementation outside?

contracts/tansu/src/contract_dao.rs
sasasamaes: Done!

contracts/tansu/src/contract_dao.rs
sasasamaes: Done. Both add_conflict_of_interest and remove_conflict_of_interest now call crate::auth_maintainers(&env, &maintainer, &project_key) directly. Renamed the parameter from caller to maintainer, and  dropped the now-redundant ConflictProposerForbidden / ConflictOnlySelf errors and their custom checks. Tests and the UI were updated to match the maintainer-only semantics.

contracts/tansu/src/contract_dao.rs
sasasamaes: Done. get_conflict_of_interest now reads from ProjectKey::ConflictOfInterest directly in the trait method — the get_conflicts helper is gone.

contracts/tansu/src/events.rs
```diff
@@ -130,3 +130,14 @@ pub struct SubProjectsUpdated {
...
     pub sub_projects: Vec<Bytes>,
 }
+
+#[contractevent]
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct ConflictOfInterestUpdated {
+    #[topic]
+    pub project_key: Bytes,
+    pub proposal_id: u32,
+    pub caller: Address,
```
tupui: We want to be clear on the naming
```suggestion
    pub maintainer: Address,
```

contracts/tansu/src/events.rs
```diff
@@ -130,3 +130,14 @@ pub struct SubProjectsUpdated {
...
+    pub added: Vec<Address>,
+    pub removed: Vec<Address>,
```
tupui: Let's simplify and use a single field `changed`

contracts/tansu/src/tests/test_dao.rs
```diff
@@ -1014,3 +1014,122 @@ fn badge_based_get_max_weight_still_works() {
...
+            &id,
+            &proposal_id,
+            &vec![&setup.env, outsider.clone()],
+        )
+        .unwrap_err()
+        .unwrap();
+    assert_eq!(err, ContractErrors::UnauthorizedSigner.into());
+
+    // Maintainer can add multiple addresses
+    let another = Address::generate(&setup.env);
```
tupui: Add something like 5 addresses (2 at first and then 3, try to add also the 1st back for instance to check that too) so that we can remove 2 (e.g. the 2nd and the 4th) and thus check the removal logic works.

contracts/tansu/src/contract_dao.rs
```diff
@@ -471,6 +471,19 @@ impl DaoTrait for Tansu {
...
 
+        // Block voters on the conflict-of-interest list
+        let conflicts: Vec<Address> = env
+            .storage()
+            .persistent()
+            .get(&types::ProjectKey::ConflictOfInterest(
+                project_key.clone(),
+                proposal_id,
+            ))
+            .unwrap_or(Vec::new(&env));
```
tupui: Use get_conflict_of_interest

contracts/tansu/src/contract_dao.rs
```diff
@@ -935,6 +948,148 @@ impl DaoTrait for Tansu {
...
+        }
+
+        let mut list = env
+            .storage()
+            .persistent()
+            .get(&types::ProjectKey::ConflictOfInterest(
+                project_key.clone(),
+                proposal_id,
+            ))
+            .unwrap_or(Vec::new(&env));
```
tupui: Use get_conflict_of_interest

contracts/tansu/src/contract_dao.rs
```diff
@@ -935,6 +948,148 @@ impl DaoTrait for Tansu {
...
+        }
+
+        let list: Vec<Address> = env
+            .storage()
+            .persistent()
+            .get(&types::ProjectKey::ConflictOfInterest(
+                project_key.clone(),
+                proposal_id,
+            ))
+            .unwrap_or(Vec::new(&env));
```
tupui: Use get_conflict_of_interest

dapp/src/components/page/proposal/ProposalTitle.tsx
```diff
@@ -145,7 +147,25 @@ const ProposalTitle: React.FC<Props> = ({
...
+                {proposal?.status == "active" && (
+                  <Button
+                    size="sm"
+                    type="secondary"
+                    onClick={() => {
+                      if (!connectedAddress) {
+                        toast.error(
+                          "Connect Wallet",
+                          "Please connect your wallet first.",
+                        );
```
tupui: Not needed. Anyone can see this without being connected.

contracts/tansu/src/contract_dao.rs
```diff
@@ -935,6 +948,148 @@ impl DaoTrait for Tansu {
...
+            added,
+            removed: Vec::new(&env),
+        }
+        .publish(&env);
+    }
+
+    /// Remove addresses from the conflict-of-interest list of a proposal.
+    ///
+    /// Only maintainers of the project can edit the list.
+    ///
```
tupui:
```suggestion
    ///
    ///
```

contracts/tansu/src/contract_dao.rs
```diff
@@ -935,6 +948,148 @@ impl DaoTrait for Tansu {
             proposal_id += 1;
         }
     }
+
+    /// Add addresses to the conflict-of-interest list of a proposal.
+    ///
+    /// Addresses on the list cannot cast a vote on the proposal. Only
+    /// maintainers of the project can edit the list.
```
tupui:
```suggestion
    /// Addresses on the list cannot cast a vote on the proposal.
```

tupui (changes requested):
Ok we are getting there.

Do you have a screenshot of the modals otherwise? I will update the contract and test the features next.

contracts/tansu/src/events.rs
sasasamaes: Done. Renamed the field to `maintainer` so it lines up with the parameter passed to the contract method.

contracts/tansu/src/events.rs
sasasamaes: Done. Collapsed `added` and `removed` into a single `changed` field. Both event publish sites now emit the affected addresses through that single name.

contracts/tansu/src/tests/test_dao.rs
sasasamaes: Done. The test now adds 2 addresses, then 3 more (including a duplicate of the first to cover the dedup branch on add) and afterwards removes the 2nd and 4th entries, so we exercise the mid-list removal path while preserving the rest.

contracts/tansu/src/contract_dao.rs
sasasamaes: Done. The vote path now reads the list through `Self::get_conflict_of_interest` instead of inlining the storage access.

contracts/tansu/src/contract_dao.rs
sasasamaes: Done. `add_conflict_of_interest` now loads the list through `Self::get_conflict_of_interest`.

contracts/tansu/src/contract_dao.rs
sasasamaes: Done. `remove_conflict_of_interest` now loads the list through `Self::get_conflict_of_interest`.

dapp/src/components/page/proposal/ProposalTitle.tsx
sasasamaes: Done. Dropped the wallet-required gate; the button now opens the modal directly. The modal already exposes the list in read-only mode for non-maintainers (edit controls only render when `canEdit` is true).

contracts/tansu/src/contract_dao.rs
sasasamaes: Done. Removed the redundant maintainer-only sentence from the `remove_conflict_of_interest` docstring.

contracts/tansu/src/contract_dao.rs
sasasamaes: Done. Trimmed the `add_conflict_of_interest` docstring to the suggested wording.

sasasamaes:
> Ok we are getting there.
> 
> Do you have a screenshot of the modals otherwise? I will update the contract and test the features next.
Captura de pantalla 2026-04-29 a las 1 33 26 p  m

/preview/conflict-modal

contracts/tansu/src/contract_dao.rs
```diff
@@ -889,6 +896,176 @@ impl DaoTrait for Tansu {
...
+    /// Proposal IDs are `0..DaoTotalProposals` for that project. It is idempotent.
+    ///
+    /// # Arguments
+    /// * `env` - The environment object
+    /// * `admin` - Admin address authorized to run migrations
+    ///
+    /// # Panics
+    /// * If `admin` is not authorized
+    /// * If migration invariants are violated
+    fn migrate_stellarpga_vote_tallies(env: Env, admin: Address) {
```
tupui: You need to merge main, this is a leftover I removed already.

contracts/tansu/src/lib.rs
```diff
@@ -165,6 +165,26 @@ pub trait DaoTrait {
     fn get_dao(env: Env, project_key: Bytes, page: u32) -> types::Dao;
 
     fn get_proposal(env: Env, project_key: Bytes, proposal_id: u32) -> types::Proposal;
+
+    fn migrate_stellarpga_vote_tallies(env: Env, admin: Address);
```
tupui: to remove as well

tupui (commented):
Nice, I will update the contract tmr and test everything and probably merge after it. Thanks!

Just something to fix now and I think we are good. I can also fix tomorrow when I pull the code to deploy the new contract.

contracts/tansu/src/contract_dao.rs
sasasamaes: Done. Removed the stale `migrate_stellarpga_vote_tallies` implementation from `contract_dao.rs` so the branch matches upstream.

contracts/tansu/src/lib.rs
sasasamaes: Done. Dropped the matching trait method from `DaoTrait` in `lib.rs`.

tupui:
Thanks for the updates. Just seeing there is a merge conflict to fix.

On my side I am pulling the contract itself and will be updating.

dapp/src/service/ReadContractService.ts
```diff
@@ -324,6 +324,23 @@ export {
  * Check whether anonymous voting is configured for a given project name.
  * Returns true when configuration exists, false otherwise.
  */
+export async function getConflictOfInterest(
```
Shadow-MMN, line 400: This function swallows every error and returns `[]`. That makes a failed contract read look identical to a real empty conflict list, which can mislead maintainers in the modal. Surfacing the error to the caller instead of collapsing it into an empty list will be a good way to improve UI.

Shadow-MMN (changes requested):
Overall code solves this issue. Great job , just pointed out some minimal change. Happy Coding ☺️.

dapp/src/service/ReadContractService.ts
sasasamaes: Done. `getConflictOfInterest` no longer swallows errors — the try/catch is gone and the contract error propagates to the caller. The modal now distinguishes a failed read from a real empty list: it shows the error message in red and exposes a Retry button (see commit 70a64cc). The misleading "anonymous voting" docstring on this function was also fixed.

sasasamaes:
Pushed two updates to address the latest round of feedback:

- **Merge conflict resolved** (commit `4af5a79`). Merged `upstream/main` into the branch. The only real conflict was in `dapp/src/components/page/proposal/ProposalTitle.tsx`, where upstream introduced the member-profile modal/`isMaintainer` simplification while this branch had already added the conflict-of-interest modal. Kept both: the member-profile flow from `main` and the new `Conflict of Interest` button/modal, sharing the cleaner inline `isMaintainer` derivation from upstream.
- **Surface errors from `getConflictOfInterest`** (commit `70a64cc`, addressing @Shadow-MMN's comment). The read service no longer swallows contract errors and returns `[]`; the error propagates to the caller. The modal now distinguishes a failed read from an empty list — it renders the error message with a Retry button instead of silently showing "No addresses have been declared." The stale docstring on that function (it was copy-pasted from the anonymous-voting helper) was fixed as well.

All dapp checks pass locally: `bun run check`, `bun run lint`, `bun run test:unit` (21 tests). PR is no longer in a conflicting state.

tupui (approved):
Thanks all, let's get this in now. I will update the contract and test locally after

tupui:
Great this works 👏 

I will make a follow up issue as we should be able to add/remove multiple addresses at the same time. Since we also already have the contract function handling lists.

tupui:
For your points @sasasamaes you might have to create a ticket on Drips' side. I am not managing this part so I don't have the visibility.

PR #96, opened 2026-04-24
Feat: Add Function to remove a vote

### LINKED ISSUE #70: Function to remove a vote
If someone is casting a vote which is malicious or does not respect some community parameter, it can be good to have a way to remove a problematic vote.

Oluwatos94:
Summary                                                                                                                                                                                                                                         - Added remove_vote to the Tansu contract callable only by a project maintainer on an active proposal within its      
voting period
  - Voter's collateral is slashed (not returned) as a penalty for malicious/non-compliant voting                          
  - Added VoteRemoved event and VoteNotFound error to the contract                                                        
  - Added full TypeScript bindings (XDR spec, type definitions, fromJSON entry)                                           
  - Added removeVoteFlow in FlowService signs and sends the remove_vote transaction                                     
  - Added RemoveVoteModal shows current voters grouped by vote type (Approve / Reject / Abstain), with a per-voter      "Remove" button                                                                                                         
  - "Remove Vote" button appears in ProposalTitle only when: proposal is active, connected user is a maintainer, and at least one vote exists                                                                                                 
  Test plan                                                                                                             
  - Connect wallet as a project maintainer                                                                                
  -  Open an active proposal that has at least one vote
  - Confirm "Remove Vote" button appears (does NOT appear for non-maintainers or proposals with no votes)                 
  - Click "Remove Vote" modal opens with voters listed by tab                                                           
  - Click "Remove" on a voter transaction is signed and sent                                                           
  - Confirm the voter no longer appears in the list and their collateral was not returned                                 
  - Confirm non-maintainer cannot call remove_vote (contract rejects with UnauthorizedSigner)                             
  - Confirm removal fails after the voting period ends (ProposalVotingTime error)   

<img width="1366" height="661" alt="Screenshot (1921)" src="https://github.com/user-attachments/assets/31ce35a9-c974-4f4a-9ae2-aa72d4b2588a" />
                                                                                                                     

closes #70


Oluwatos94:
Hi @tupui, This PR is ready for review

tupui:
Yes thanks I've seen :) just many things right now. I will try to make a review soon

contracts/tansu/src/contract_dao.rs
```diff
@@ -350,6 +350,118 @@ impl DaoTrait for Tansu {
...
+                    project_key.clone(),
+                    vec![&env, 0u128, 0u128, 0u128],
+                    vec![&env, 0u128, 0u128, 0u128],
+                );
+                let mut rebuilt = types::VoteTallies::AnonymousVote(zero_agg);
+                for v in remaining_votes.iter() {
+                    update_proposal_tallies(&env, &mut rebuilt, &v);
+                }
+                proposal_tallies = rebuilt;
+            }
```
tupui: We cannot do this. It will be too expensive with even 10 voters. This is why I added the tallies now. Instead we must do a subtraction of the voter's commitments. This is a bit more complex as we need to do 3 subtractions as we have 3 commitments per vote.

contracts/tansu/src/contract_dao.rs
```diff
@@ -350,6 +350,118 @@ impl DaoTrait for Tansu {
...
+            Ok(Some(proposal)) => proposal,
+            _ => panic_with_error!(&env, &errors::ContractErrors::NoProposalorPageFound),
+        };
+
+        if proposal.status != types::ProposalStatus::Active {
+            panic_with_error!(&env, &errors::ContractErrors::ProposalActive);
+        }
+        if env.ledger().timestamp() >= proposal.vote_data.voting_ends_at {
+            panic_with_error!(&env, &errors::ContractErrors::ProposalVotingTime);
+        }
```
tupui: we don't want to have that check because we would notice issues maybe after the voting period ends. So we want to be able to remove a bad vote at that time.

contracts/tansu/src/tests/test_dao.rs
```diff
@@ -1014,3 +1016,190 @@ fn badge_based_get_max_weight_still_works() {
...
+    // Proposal can still execute; the removed vote is not counted
+    setup.env.ledger().set_timestamp(voting_ends_at + 1);
+    let result = setup
+        .contract
+        .execute(&setup.grogu, &id, &proposal_id, &None, &None);
+    assert_eq!(result, ProposalStatus::Cancelled);
+}
+
+#[test]
+fn remove_vote_fails_for_non_maintainer() {
```
tupui: Let's put this in the same test above, just try to remove the vote first with a non maintainer.

contracts/tansu/src/tests/test_dao.rs
```diff
@@ -1014,3 +1016,190 @@ fn badge_based_get_max_weight_still_works() {
...
+
+    let outsider = Address::generate(&setup.env);
+    let result = setup
+        .contract
+        .try_remove_vote(&outsider, &id, &proposal_id, &setup.mando);
+    assert_eq!(result, Err(Ok(ContractErrors::UnauthorizedSigner.into())));
+}
+
+#[test]
+fn remove_vote_fails_when_vote_not_found() {
```
tupui: Same with this one.

contracts/tansu/src/tests/test_dao.rs
```diff
@@ -1014,3 +1016,190 @@ fn badge_based_get_max_weight_still_works() {
...
+
+    let non_voter = Address::generate(&setup.env);
+    let result = setup
+        .contract
+        .try_remove_vote(&setup.grogu, &id, &proposal_id, &non_voter);
+    assert_eq!(result, Err(Ok(ContractErrors::VoteNotFound.into())));
+}
+
+#[test]
+fn remove_vote_fails_after_voting_period() {
```
tupui: To remove

contracts/tansu/src/tests/test_dao.rs
tupui, line 1: We need to actually test that the voting results make sense in terms of tallies. Like we need to make a few votes, and see that the outcome of the vote would change if we remove a specific vote. The voting power here matters, not just the number of votes. We need this with public and anonymous votes as the maths is different and slightly tricky on the anonymous part with the commitment handling.

tupui (changes requested):
Thanks, I had some time to review and have some things to look into.

Oluwatos94:
@tupui Thanks for the reviews. All reviews acknowledged and changes would be made soon.

tupui:
@Oluwatos94 I also tried to check the UI but do not see the new components on a live proposal. Please make sure to test locally and ideally post a screenshot here.

Oluwatos94:
> @Oluwatos94 I also tried to check the UI but do not see the new components on a live proposal. Please make sure to test locally and ideally post a screenshot here.

Okay, would do that. The UI is only visible to Maintainer of a project.

tupui:
Just to be clear, I did try being a maintainer of the project. I can see the "eye" for the anonymous voting part if logged as a maintainer for instance.

Oluwatos94:
> Just to be clear, I did try being a maintainer of the project. I can see the "eye" for the anonymous voting part if logged as a maintainer for instance.

Have you tried to redeploy contract? I've provided a screenshot above.

contracts/tansu/src/contract_dao.rs
```diff
@@ -350,6 +350,125 @@ impl DaoTrait for Tansu {
...
+                let bls12_381 = env.crypto().bls12_381();
+                // BLS12-381 scalar field order r. Multiplying by (r - weight) ≡ -weight (mod r).
+                let group_order = U256::from_parts(
+                    &env,
+                    0x73eda753299d7d48u64,
+                    0x3339d80809a1d805u64,
+                    0x53bda402fffe5bfeu64,
+                    0xffffffff00000001u64,
+                );
+                let neg_weight = group_order.sub(&U256::from_u32(&env, vote_choice.weight));
```
tupui: This looks convoluted. In `test_anonymous_votes.rs` we have this

```rust
fn neg_value(env: &Env, value: G1Affine) -> G1Affine {
    // Create -1 in Fr field using fr_sub(0, 1)
    let bls12_381 = env.crypto().bls12_381();
    let zero = U256::from_u32(env, 0);
    let one = U256::from_u32(env, 1);
    let neg_one = bls12_381.fr_sub(&zero.into(), &one.into());

    // negate the value
    bls12_381.g1_mul(&value, &neg_one)
}
```

I expected something clearer like this.

dapp/src/components/page/proposal/ProposalTitle.tsx
Shadow-MMN, line 1: The current implementation in this file relies on `totalVotes > 0` as the only condition for showing the Remove Vote action. This assumption breaks for anonymous proposals.

In `utils.ts` , anonymous proposal vote data is intentionally zeroed out. Additionally, `RemoveVoteModal.tsx` depends on `voteStatus[currentType].voters`, which is not populated for anonymous votes. As a result, the feature is effectively unavailable for anonymous proposals without any explicit indication in the UI. The current behavior creates a silent inconsistency between UI availability and actual functionality.

We can either explicitly restrict the Remove Vote action to public proposals in the UI, or introduce a maintainer-only data source that allows access to anonymous voter data for removal.

Shadow-MMN (changes requested):
Overall code looks solid but when tested I couldn't see the UI changes you made.

Image

please have a look at it

dapp/src/components/page/proposal/ProposalTitle.tsx
tupui: Thanks for pointing this out. Since this is a maintainer only action, I am also ok having a non-complicated logic and always show the button if that helps simplify the code and maintainability. (Otherwise in `proposal.vote_data.votes` we always have all votes we can inspect. We have a list of `Vote` which is a `Vec` of `AnonymousVote` or `PublicVote`)

You both know now that I prefer to have non-complex code and smaller, cleaner is my preference 😉

Oluwatos94: @Shadow-MMN thanks for the review, from the screenshot in the PR the remove vote is only visible when the vote is > 0 reason you're not seeing the UI component from your test.

@tupui Yeah, I understand. I'd look into it and make more cleaner. Thanks!

Shadow-MMN: > @Shadow-MMN thanks for the review, from the screenshot in the PR the remove vote is only visible when the vote is > 0 reason you're not seeing the UI component from your test.
> 
> @tupui Yeah, I understand. I'd look into it and make more cleaner. Thanks!

oh my bad , will check it out after making a vote now

Oluwatos94: I'm making an update soon that would always show regardless of whether voted or not. Since it's only visible to the maintainer of the project. I think it nice to keep it simple as suggested by Tupui

Shadow-MMN: > I'm making an update soon that would always show regardless of whether voted or not. Since it's only visible to the maintainer of the project. I think it nice to keep it simple as suggested by Tupui

Yh go for it 💪🏾

Shadow-MMN:
Screenshot 2026-05-01 at 01 37 01

Correct me if I am wrong but I think the maintainer should not be able to remove this cancelled vote that was voted by him during the proposal creation. If possible can we add the copy button modal so that the address can be copied.
Excellent work so far .

Shadow-MMN:
> Screenshot 2026-05-01 at 01 37 01
> Correct me if I am wrong but I think the maintainer should not be able to remove this cancelled vote that was voted by him during the proposal creation. If possible can we add the copy button modal so that the address can be copied. Excellent work so far .

Screenshot 2026-05-01 at 01 42 44

This is what I got when I tried removing it. I think it is better to hide the remove for that specific vote

tupui:
> 
> 
> Screenshot 2026-05-01 at 01 37 01
> 
> 
> 
> Correct me if I am wrong but I think the maintainer should not be able to remove this cancelled vote that was voted by him during the proposal creation. If possible can we add the copy button modal so that the address can be copied.
> 
> Excellent work so far .

It's ok. A maintainer has full authority and now anyone can make proposals, not just maintainers. Let's not complicate things.

If I want to guard that later I will do on the contract side.

tupui:
Also @Shadow-MMN it's expected not to work unless you deploy your own contract as I did not yet upgrade.

The conflict of interest PR is going to go in first. Then this one for the contract.

Shadow-MMN:
> > Screenshot 2026-05-01 at 01 37 01
> > Correct me if I am wrong but I think the maintainer should not be able to remove this cancelled vote that was voted by him during the proposal creation. If possible can we add the copy button modal so that the address can be copied.
> > Excellent work so far .
> 
> It's ok. A maintainer has full authority and now anyone can make proposals, not just maintainers. Let's not complicate things.
> 
> If I want to guard that later I will do on the contract side.

Ok

tupui (approved):
Thanks again for that. Getting it in 🚀

tupui:
@Oluwatos94 @Shadow-MMN  Please check the deployment. This is not working in the UI I do not see votes to remove.

Oluwatos94:
> @Oluwatos94 @Shadow-MMN Please check the deployment. This is not working in the UI I do not see votes to remove.

Ok, I will look into it as soon as possible.

PR #101, opened 2026-04-26
feat(ipfs): implement multi-provider pinning with retries

### LINKED ISSUE #39: Dual pinning on IPFS

Richiey1:
This PR enhances the IPFS pinning logic to ensure maximum content availability.

Key changes:
- Added support for Pinata as a third pinning provider alongside Storacha and Filebase.
- Implemented  logic to parallelize pinning across multiple services.
- Added a robust  helper with exponential backoff for transient pinning failures.
- Updated all DApp upload flows to utilize the new redundant pinning system.
- Added comprehensive unit tests in  (4/4 passed).

Closes #39

### DIFF
--- a/dapp/.env.example
+++ b/dapp/.env.example
@@ -13,3 +13,4 @@ PUBLIC_DELEGATION_API_URL="https://ipfs-testnet.tansu.dev"
 
 # Dual IPFS pinning — Filebase backup (optional but recommended for redundancy)
 FILEBASE_API_KEY=""
+PINATA_JWT=""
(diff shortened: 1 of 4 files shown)


Richiey1:
Hi @Consulting-Manao, I've implemented the multi-provider IPFS pinning with robust retries as per issue #39. Ready for review!

tupui:
Not sure what you are trying to do here... @Richiey1

PR #103, opened 2026-04-26
feat: add rich content editor with image upload to generic project README

### LINKED ISSUE #65: Rich content in Readme for generic projects
When we use generic projects, we should also be able to fill out rich information with images and tables, etc. We already have some logics to do this when filling out proposals. It should work the same way with storage on IPFS in the same manner.

Codekill33:
Replace plain textarea with SimpleMarkdownEditor and image upload controls
(same pattern as proposals) so generic project READMEs support images,
tables, and other rich markdown. Images are stored alongside README.md in
IPFS; relative paths are rewritten to absolute gateway URLs on display.

Closes #65


tupui (changes requested):
@Codekill33 seems like you did not test it, I don't see the readme being rendering. See https://deploy-preview-103--staging-tansu.netlify.app/project/?name=stellarpga

tupui:
@Codekill33 friendly reminder.

Codekill33:
Alright, will do that shortly.

tupui (changes requested):
This is not what I asked for.

On a non-generic project like here
https://deploy-preview-103--staging-tansu.netlify.app/project/?name=tansu we should not change anything. We have button to show the readme and this is working good as expected and we do not want to see the readme elsewhere.

For generic projects, https://deploy-preview-103--staging-tansu.netlify.app/project/?name=stellarpga this is now good. But also not asked here. Instead this is what we are doing in #122

This one is just about creating and editing rich readmes.

I did not check the code yet. Just be sure to have minimal changes and only relevant for this work.

Codekill33:
Hope it's okay now?

tupui:
@Codekill33 I merged #122 please make sure this works properly with it. I am expecting that you create yourself a project and check that you can also update. You can do all that locally and post screenshots here as proof. See #122 this is what they did there for instance.

dapp/src/components/page/dashboard/CreateProjectModal.tsx
```diff
@@ -369,10 +373,21 @@ ${maintainerGithubs.map((gh) => `[[PRINCIPALS]]\ngithub="${gh}"`).join("\n\n")}
...
-          type: "text/plain",
+        let finalReadme = readmeContent || "";
+        const imageFilesToInclude: File[] = [];
+        readmeImageFiles.forEach((img) => {
+          if (finalReadme.includes(img.localUrl)) {
+            finalReadme = finalReadme.replace(
+              new RegExp(img.localUrl.replace(/[.*+?^${}()|[\]\\]/g, "\\$&"), "g"),
+              img.publicUrl,
+            );
+            imageFilesToInclude.push(new File([img.source], img.publicUrl));
```
Shadow-MMN: The image attachments are recreated with `new File([img.source], img.publicUrl)` but the MIME type is dropped. That means the saved files no longer carry image/png, image/jpeg, etc., which can cause inconsistent handling when the files are packed and uploaded. Can we consider preserving the type when rebuilding the file. E.g ` new File([img.source], img.publicUrl, { type: img.source.type })`.

dapp/src/components/page/dashboard/CreateProjectModal.tsx
```diff
@@ -854,15 +869,120 @@ ${maintainerGithubs.map((gh) => `[[PRINCIPALS]]\ngithub="${gh}"`).join("\n\n")}
...
+                                );
+                                return;
+                              }
+                              if (file.size > 5 * 1024 * 1024) {
+                                setReadmeImageError(
+                                  "Please upload an image smaller than 5MB",
+                                );
+                                return;
+                              }
+                              const localUrl = URL.createObjectURL(file);
```
Shadow-MMN: This is currently only revoked when a user manually removes an attachment. If the modal is closed while attachments are still present, the associated blob URLs are not cleaned up, leading to potential memory leaks.

In `UpdateConfigModal`, this issue is more pronounced because the attachment state is reset on reopen (line 294) without revoking previously created object URLs.

Suggested change: implement a cleanup routine on modal close or component unmount that iterates over stored localUrl values and calls URL.revokeObjectURL before clearing the state.

Shadow-MMN (changes requested):
The overall code looks solid. Just view issues I pointed out. Happy Coding ☺️.

tupui:
@Codekill33 do you intend to finish? You would still get points if we merge before the end of the week. Up to you otherwise I close.

Codekill33:
Yes i intend to finish i  had a medical emergency, will work on it.

tupui:
Sorry to read that. Hope all is well.

Codekill33:
All is well now. Will fix issue soon please bear with me.

EslaSk:
Screenshot 2026-05-06 095045
Screenshot 2026-05-06 095054
Screenshot 2026-05-06 095522
Screenshot 2026-05-06 095529
Screenshot 2026-05-06 095555

Tested on staging — created a generic project with a rich markdown README and updated it successfully. Both the Edit and Preview tabs work as expected

tupui:
Thanks for the updates! @Shadow-MMN this seems to be ready for a review if you have time.

Shadow-MMN:
> Thanks for the updates! @Shadow-MMN this seems to be ready for a review if you have time.

On it now

dapp/src/components/page/dashboard/CreateProjectModal.tsx
```diff
@@ -76,6 +80,13 @@ const CreateProjectModal: FC<ModalProps> = ({ onClose }) => {
   const walletKey = useStore(connectedPublicKey);
   const isWalletReady = useStore(walletInitialized);
 
+  // Revoke blob URLs on unmount to avoid memory leaks
```
Shadow-MMN: `readmeImageFiles` blob URLs are revoked on every state change, which breaks existing attachments as soon as a second image is added or one is removed. In both create and update flows, the cleanup effect runs whenever the array changes, not just on unmount, so earlier `blob:` URLs get invalidated while they are still referenced by the editor preview and markdown. This is a functional regression for multi-image READMEs.
Can we consider revoking the a`blob` url only when they are no longer needed and not on every state update.

dapp/src/components/page/project/UpdateConfigModal.tsx
```diff
@@ -290,6 +301,12 @@ const UpdateConfigModal = () => {
   useEffect(() => {
     if (!open) return;
 
+    setReadmeImageFiles((prev) => {
```
Shadow-MMN, line 331: This clears `readmeImageFiles` on open and only uploads the images currently tracked in that state, but it never reconstructs existing image attachments from the `README` fetched from `IPFS`. That means a user who edits text and saves can silently orphan image files already referenced by the `README`. Can we consider preserving existing `README` image references during updates or reconstructing them into the tracked attachment state before upload.

Shadow-MMN (changes requested):
Nice work on your changes. I have pointed out some areas that needs changes and again don't forget to update your branch , a pr was recently merged. Happy coding 😊.

dapp/src/components/page/project/UpdateConfigModal.tsx
Codekill33: Fixed — on submit, the code now parses the README for any existing relative image references and re-fetches them from the old IPFS CID, including them in the new CAR package so they are not orphaned after the update.

dapp/src/components/page/dashboard/CreateProjectModal.tsx
Codekill33: Fixed — replaced the [readmeImageFiles] dependency with a ref (readmeImageFilesRef) that stays in sync with the latest state. The cleanup effect now uses an empty dependency array so it only runs on unmount, not on every state change.

tupui (commented):
Still looks like we are duplicating the code and not reusing anything no?

We already should have code to create rich content, why are we re-doing all the work now twice?

EslaSk:
I'm a bit confused

tupui:
@Codekill33 what are you confused with?

EslaSk:
With your last reply, need more clarity.

tupui:
@Codekill33 look at the changes, to me it looks like we duplicate some code.

We also already have some existing code to create a rich markdown document when we create proposals. So I would expect that there would be some code we can share. No?

EslaSk:
@tupui You're right, I'll extract the image upload logic into a shared component and reuse it across CreateProposalModal, CreateProjectModal, and UpdateConfigModal. That should do it, yes?

tupui:
Yes, I expect more line deleted than added basically 😅

Shadow-MMN (changes requested):
Nice work on what you did but the shared `MarkdownEditorWithImages` cleanup is too aggressive. It revokes `blob:` URLs on unmount, but `UpdateConfigModal` unmounts the editor during normal step navigation, so going Back then Next leaves the `README` markdown pointing at revoked URLs and the previews break.
Can we consider moving the `object-URL` cleanup to the modal close/unmount lifecycle instead of the editor component, then re-test the `Back/Next` flow in `UpdateConfigModal` and the same editor in `CreateProjectModal`

EslaSk:
Alright, on it.

Shadow-MMN (changes requested):
`README` content is rendered directly, but relative image and link paths are not rewritten to absolute gateway/repository `URLs` before display. That means assets and intra-repo links can still break depending on the `README` structure. We should normalize those paths during render so the editor behavior matches the PR description.

Then the `CreateProjectModal.tsx` is all red in linting please have a look at it. The `MarkdownEditorWithImages.tsx` needs formatting cleanup for lint. 

Once we are done with this then we can be ready to merge. Nice work so far and Happy coding ☺️.

Shadow-MMN (changes requested):
Excellent work, the `README` editor preview now rewrites relative paths correctly, but the read-only README render path still doesn’t. In `ReadmeViewer / ReadMoreModal`, relative image and link URLs are still rendered as-is, so `README` assets and intra-repo links can break after display. We should apply the same normalization there too.

Shadow-MMN (changes requested):
Every blocking issues have been resolved. Kindly run `cd dapp` and `bun run format` to format the `MarkdownEditorWithImage.tsx` file and push the changes then we are good to go. Happy Coding ☺️.

EslaSk:
Alright, on it

dapp/.env.example
Shadow-MMN, line 1: Why are we deleting this file ?. The last commit was right any need to delete this file ?

Shadow-MMN (changes requested):
The last commit before this was actually better was about to approve it after testing. Can we go back to it i.e commit f3fd19f

Shadow-MMN (approved):
Everything looks good in this commit. I tested the changes locally and all the issues I previously raised have been resolved properly. Nice work on the fixes.

@tupui I think this is safe to merge, though feel free to check if there’s anything I may have missed. The PR looks ready from my side.

Thank you @Codekill33 for contributing to Tansu. Happy coding ☺️

EslaSk:
Apologise for that, when I ran git add -a to stage the formatting changes,
it accidentally picked up the .env.example deletion as well. I have
reverted it.

EslaSk:
Thank you for giving me a chance.

Shadow-MMN:
> Author

No problem , you have resolved it. Nice work 💪🏾.

tupui (approved):

PR #114, opened 2026-04-28
fix: removed default approve vote selection

### LINKED ISSUE #105: No default for vote casting
We must not pre-select a voting outcome when casting a vote. Right now we default to the following:

Image

Nothing should be selected.

ComputerOracle:
Closes #105

---

This fixes the issue where "Approve" is selected by default when opening the voting modal.

The initial state was set to VoteType.APPROVE, which caused the option to be pre-selected. I changed it to null so no option is selected when the modal opens.

Now the user has to actively choose a vote option, which keeps the behavior neutral and matches the expected flow.

### DIFF
--- a/dapp/src/components/page/proposal/VotingModal.tsx
+++ b/dapp/src/components/page/proposal/VotingModal.tsx
@@ -25,9 +25,7 @@ const VotingModal: React.FC<VotersModalProps> = ({
   setIsVoted,
   onClose,
 }) => {
-  const [selectedOption, setSelectedOption] = useState<VoteType | null>(
-    VoteType.APPROVE,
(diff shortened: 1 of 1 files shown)


tupui (changes requested):
Thanks, so this is not working as expected. Now all is selected. You can check here and try to vote yourself

https://deploy-preview-114--staging-tansu.netlify.app/proposal/?id=6&name=tansu
Screenshot 2026-04-29 at 10 38 29

tupui:
I will close the other one then. Please don't open new PRs for the same work in the future. You can update the working branch and force push if needed. Or just merge main. Many options.

PR #116, opened 2026-04-28
Fix maintainer projects in community engagements

### LINKED ISSUE #110: Missing projects in community engagement view
On this project, https://testnet.tansu.dev/project/?name=stellarpga we have GBMOFLBLOINEUB6BTAF6U3PY6YYYMPZ5LT7R7VQCR6RXALBJZOJ4IYTW as a maintainer but as we can see on the membership profile, it does not show that project `stellarpga`

Image

OlaBakare:
The community engagement modal only checked the first page of projects for maintainer roles, so maintainer-only projects on later pages, like `stellarpga`, were missing from the profile.
I added a `getAllProjects()` reader that paginates through all project pages, then changed `MemberProfileModal` to use it when checking whether the member is a maintainer, with canonical project-key derivation for deduping entries.

closes #110

### DIFF
--- a/dapp/src/components/page/dashboard/MemberProfileModal.tsx
+++ b/dapp/src/components/page/dashboard/MemberProfileModal.tsx
@@ -11,13 +11,14 @@ import { connectedPublicKey } from "../../../utils/store";
 import { refreshLocalStorage } from "@service/StateService";
 import {
   getProjectFromId,
-  getProjectsPage,
+  getAllProjects,
 } from "../../../service/ReadContractService";
 import { navigate } from "astro:transitions/client";
 import { Buffer } from "buffer";
 import OnChainActions from "./OnChainActions";
 import { badgeName } from "../../../utils/badges";
 import AddressDisplay from "../proposal/AddressDisplay"; // use existing component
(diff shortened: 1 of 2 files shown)


tupui (changes requested):
Oh I see the issue. Thanks for working on this. Well in that case this is going to be a big problem in the future if we get many projects. Then it's going to have to load too many projects one by one. In that case, can you remove the maintainer indication and only keep the badges to be shown in the profiles? We have to do this until I rethink the logic to have a way to list all maintainers/addresses in a more efficient way.

OlaBakare:
Thanks for the review. I understand the concern about loading too many projects individually.

I will update the PR to remove the maintainer indication logic and keep only the badges that are already shown on the profile, as requested.

OlaBakare:
I have updated the PR based on your request.

The maintainer indication logic and all-project pagination check have been removed. The profile now only keeps the existing badges display, without trying to detect maintainer projects by loading all projects.

Please review again when you have time.

tupui (approved):
Thank you, this works as I expect now 👍

PR #117, opened 2026-04-29
Fix/loading keyfile

### LINKED ISSUE #106: Loading indicator on anonymous vote decoding
When decoding large ballots, it can take a few seconds after loading the key file. We are stuck on seeing this while this happens in the background

Image

We should instead have a loading indicator (reuse existing loading components like the one we have when loading all proposals.)

ComputerOracle:
This adds a loading state when processing the key file in the voting modal.

Large ballots take a few seconds to decode, and before this change the UI looked stuck during that time. I added a simple loading message so the user knows the process is ongoing.

Once processing is complete, the flow continues as before.

closes #106

### DIFF
--- a/dapp/src/components/page/proposal/VerifyAnonymousVotesModal.tsx
+++ b/dapp/src/components/page/proposal/VerifyAnonymousVotesModal.tsx
@@ -30,7 +30,9 @@ const VerifyAnonymousVotesModal: React.FC<Props> = ({
     null,
   );
   const [decodedVotes, setDecodedVotes] = useState<DecodedVote[]>([]);
+  const [isProcessing, setIsProcessing] = useState(false);
 
+  const [isProcessing, setIsProcessing] = useState(false);
   const computeTalliesAndProof = async (privKey: string) => {
     try {
       const data = await computeAnonymousVotingData(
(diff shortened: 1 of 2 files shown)


dapp/src/components/page/proposal/VotingModal.tsx
tupui, line 1: No unnecessary changes. You are using the wrong branch. You must start new PRs from main instead of using your other branches.

tupui (changes requested):

dapp/src/components/page/proposal/VerifyAnonymousVotesModal.tsx
```diff
@@ -30,7 +30,9 @@ const VerifyAnonymousVotesModal: React.FC<Props> = ({
     null,
   );
   const [decodedVotes, setDecodedVotes] = useState<DecodedVote[]>([]);
+  const [isProcessing, setIsProcessing] = useState(false);
 
+  const [isProcessing, setIsProcessing] = useState(false);
```
tupui, line 35: This is causing a build issue I think

tupui (changes requested):

tupui:
You opened #123 now instead of updating this... I am closing this one then.

PR #118, opened 2026-04-29
Fix/highlight exceeded weight

### LINKED ISSUE #108: Anonymous votes inconsistency flagging
When decoding anonymous votes, we are getting the weight and max allowed weight for someone. In some very specific instances, the weight can go above the allowed value (vote casted before update of the voting power). When that happens like we see here, we must highlight the row so that we can inspect the data if we want. This can be all fine and legitimate so no blocking or anything must happen.

Image

ComputerOracle:
Adds a visual highlight for rows where the vote weight exceeds the max allowed weight.

This helps make those edge cases easier to spot while reviewing decoded votes. The check handles both numeric and string values and ignores cases like "N/A".

The change is purely visual and does not affect any validation or flow.

closes #108

### DIFF
--- a/dapp/src/components/page/proposal/VerifyAnonymousVotesModal.tsx
+++ b/dapp/src/components/page/proposal/VerifyAnonymousVotesModal.tsx
@@ -30,7 +30,9 @@ const VerifyAnonymousVotesModal: React.FC<Props> = ({
     null,
   );
   const [decodedVotes, setDecodedVotes] = useState<DecodedVote[]>([]);
+  const [isProcessing, setIsProcessing] = useState(false);
 
+  const [isProcessing, setIsProcessing] = useState(false);
   const computeTalliesAndProof = async (privKey: string) => {
     try {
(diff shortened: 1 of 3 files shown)


dapp/src/components/page/proposal/VerifyAnonymousVotesModal.tsx
```diff
@@ -30,7 +30,9 @@ const VerifyAnonymousVotesModal: React.FC<Props> = ({
     null,
   );
   const [decodedVotes, setDecodedVotes] = useState<DecodedVote[]>([]);
+  const [isProcessing, setIsProcessing] = useState(false);
 
+  const [isProcessing, setIsProcessing] = useState(false);
```
tupui, line 35: On this one too.

tupui (changes requested):
Same issue as in some other PR. You must really use main for all your different PRs otherwise we will have diff issues.

tupui:
You did this again. I am crying!

PR #119, opened 2026-04-29
fix: make proposal creator address clickable

### LINKED ISSUE #109: Wrong label on proposal creator role
There is a wrong labelling on the role proposal creator have in a project. Behind the scenes this is because that label checks that the connected address is a maintainer. This is wrong. Instead, we don't want to show any label but have the address clickable like we have everywhere else. This would let us open the membership model where we can see projects roles.

Image

nice-bills:
Closes #109 - Wrong label on proposal creator role
Problem
The proposal creator address was incorrectly showing a "maintainer" badge. This was wrong because the code was checking if the connected wallet was a maintainer, not showing info about the proposal creator.
Solution
- Removed the incorrect maintainer badge that was based on the viewer's wallet
- Made the proposal creator address clickable
- Clicking opens MemberProfileModal where users can view the creator's actual project roles
Changes
- dapp/src/components/page/proposal/ProposalTitle.tsx - Main fix
- dapp/src/components/page/proposal/AnonymousTalliesDisplay.tsx - Prettier formatting
- dapp/src/components/page/proposal/ExportDecodedVotesModal.tsx - Prettier formatting

Before 
image

After
Screenshot_29-Apr_07-49-52_5163

Screenshot_29-Apr_07-50-18_25380

@tupui, please let me know if there are any changes you would like for me to make during your review

### DIFF
--- a/dapp/src/components/page/proposal/ProposalTitle.tsx
+++ b/dapp/src/components/page/proposal/ProposalTitle.tsx
@@ -1,9 +1,12 @@
 import { useStore } from "@nanostores/react";
 import Button from "components/utils/Button";
-import { useEffect, useState } from "react";
+import { useState } from "react";
+import type { Member } from "packages/tansu";
 import type { ProposalView } from "types/proposal";
 import { connectedPublicKey } from "utils/store";
 import { toast, truncateMiddle } from "utils/utils";
+import { getMember } from "@service/ReadContractService";
+import MemberProfileModal from "components/page/dashboard/MemberProfileModal";
 import ProposalStatusSection from "./ProposalStatusSection";
 import VoteStatusBar from "./VoteStatusBar";
 import VotingResultModal from "./VotingResultModal";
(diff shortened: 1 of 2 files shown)


tupui (commented):
Nice, seems to be working. @Shadow-MMN would you mind having a quick look at the code?

Shadow-MMN:
I am just seeing this now , will have a look.

dapp/src/components/page/proposal/ProposalTitle.tsx
```diff
@@ -63,11 +42,22 @@ const ProposalTitle: React.FC<Props> = ({
     setShowVotingResultModal(true);
   };
 
+  const openMemberProfile = async () => {
```
Shadow-MMN, line 46: Wouldn't it be better to have a try catch block inside an async function ?. Just incase there is an error and the getMember didn't work as espected. A fallback UI would be nice for the catch block instead of silent failures if the fetch fails

dapp/src/components/page/proposal/ProposalTitle.tsx
```diff
@@ -92,18 +82,14 @@ const ProposalTitle: React.FC<Props> = ({
             <div className="flex items-center gap-3">
               <p className="leading-4 text-base text-[#695A77]">Created by</p>
               <div className="flex items-center gap-2">
-                <p className="leading-4 text-base font-semibold text-primary font-mono">
+                <p
```
Shadow-MMN, line 93: It may be worth adding a loading state when fetching the member profile. Currently, there is no feedback between click and modal display.

Shadow-MMN (changes requested):
I have tested everything and it works as expected . Nice work but I have pointed out areas that needs a little improvement, once done it will be ready to merge. Happy coding ☺️.

For the loading we can use the spinner already that is already integrated in the codebase to improve ux.

dapp/src/components/page/proposal/ProposalTitle.tsx
nice-bills: thanks for the feedback. making the needed changes now

Shadow-MMN (approved):
Nice improvement, you addressed all the issues I mentioned. Good job. Happy Coding ☺️.

@tupui LGTM

dapp/src/components/page/proposal/ProposalTitle.tsx
```diff
@@ -92,18 +90,21 @@ const ProposalTitle: React.FC<Props> = ({
...
+                  className={`leading-4 text-base font-semibold text-primary font-mono ${
+                    isLoadingProfile
+                      ? "opacity-50 cursor-wait"
```
tupui:
```suggestion
                      ? "opacity-50"
```

dapp/src/components/page/proposal/ProposalTitle.tsx
```diff
@@ -92,18 +90,21 @@ const ProposalTitle: React.FC<Props> = ({
...
+                  }`}
+                  onClick={isLoadingProfile ? undefined : openMemberProfile}
+                >
                   {proposal?.proposer
                     ? truncateMiddle(proposal.proposer, 20)
                     : ""}
+                  {isLoadingProfile && (
+                    <span className="ml-2 inline-block w-3 h-3 border-2 border-[#311255] border-t-transparent rounded-full animate-spin" />
+                  )}
                 </p>
```
tupui:
```suggestion
: ""}
</p>
```

tupui (commented):
Thanks both. I am just removing the triple spinners to just keep the single one we have everywhere for consistency.

PR #121, opened 2026-04-29
fix: avoid preselecting vote option

### LINKED ISSUE #105: No default for vote casting

ComputerOracle:
Fixes issue where a vote option was pre-selected in the voting modal.

The state now initializes as null so no option is selected by default. The checkbox component was updated to explicitly handle null as an unselected state while preserving existing behavior.

This keeps the logic simple and avoids using any workaround values.

closes #105

### DIFF
--- a/dapp/src/components/page/proposal/VoteTypeCheckbox.tsx
+++ b/dapp/src/components/page/proposal/VoteTypeCheckbox.tsx
@@ -23,11 +23,7 @@ const VoteTypeCheckbox: FC<Props> = ({
   const width = `${sizeMap[size]}px`,
     height = `${sizeMap[size]}px`;
   const renderIcon = () => {
-    if (currentVoteType === null) {
-      return <img src="/icons/check-blank.svg" style={{ width, height }} />;
-    }
-
-    if (voteType == currentVoteType || currentVoteType === undefined) {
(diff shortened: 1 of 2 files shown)


dapp/src/components/page/proposal/VoteTypeCheckbox.tsx
```diff
@@ -23,7 +23,11 @@ const VoteTypeCheckbox: FC<Props> = ({
   const width = `${sizeMap[size]}px`,
     height = `${sizeMap[size]}px`;
   const renderIcon = () => {
-    if (voteType == currentVoteType || !currentVoteType) {
+    if (currentVoteType === null) {
+      return <img src="/icons/check-blank.svg" style={{ width, height }} />;
+    }
+
+    if (voteType == currentVoteType || currentVoteType === undefined) {
```
tupui: Is it then not enough to just do this?

```suggestion
    if (voteType == currentVoteType) {
```

Why do we check for null and undefined? We are already having a default at the end of the function that returns `check-blank.svg`

tupui (approved):
That works, thanks!

PR #122, opened 2026-04-29
feat: Markdown rendering of generic project readme

### LINKED ISSUE #83: Markdown rendering of generic project readme
Right now we do not render the markdown for generic project's readme. E.g. https://testnet.tansu.dev/project/?name=stellarpga

Image

We should do as we do for normal projects when we click on the readme button.

E.g. see https://testnet.tansu.dev/project/?name=tansu

Ideally we should reuse components as we want to do something similar in terms of rendering.

Oluwatos94:
Summary

Generic (non-SOFTWARE) projects like stellarpga already fetch their README from IPFS into configData.readmeContent, but the content was never rendered it was effectively invisible to users.
                                                                                                                          
- Show the Read More button for all project types, not just SOFTWARE projects                                                                                                                                                                     
- Pass the pre-fetched readmeContent into ReadMoreModal for non-SOFTWARE projects so it uses the IPFS content
  directly, without an extra GitHub fetch                                                                                 
- Hide the Clone Repository and Go to Releases actions inside the modal for non-SOFTWARE projects (those actions are
  GitHub-specific and don't apply)                                                                                        
                                                                                                                          
The markdown rendering itself reuses the existing markdown-to-jsx setup already in ReadMoreModal, so no new rendering   logic was introduced.   

<img width="1366" height="669" alt="Screenshot (1922)" src="https://github.com/user-attachments/assets/39210998-f8d5-46b4-baed-27d4ddf18399" />


closes #83

### DIFF
--- a/dapp/src/components/page/project/CommitHistory.jsx
+++ b/dapp/src/components/page/project/CommitHistory.jsx
@@ -1,4 +1,5 @@
 import { useStore } from "@nanostores/react";
+import Markdown from "markdown-to-jsx";
 import { useEffect, useState } from "react";
 import { getCommitHistory } from "../../../service/GithubService.ts";
 import {
@@ -23,7 +24,6 @@ const CommitHistory = () => {
   const [currentPage, setCurrentPage] = useState(1);
   const [isLoading, setIsLoading] = useState(true);
   const [loadError, setLoadError] = useState(null);
-  const [readmeExpanded, setReadmeExpanded] = useState(false);
 
   const fetchCommitHistory = async (page = 1) => {
     setLoadError(null);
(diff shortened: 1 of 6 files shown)


tupui:
I cross linked some work that would allow to add images to the readme. Will this work with it too?

Oluwatos94:
> I cross linked some work that would allow to add images to the readme. Will this work with it too?

I doubt, let me cross check and make adjustment to allow it.

Oluwatos94:
@tupui , I've updated it to work with the linked PR

tupui:
Can you double check @Oluwatos94 here I don't see the readme being rendered https://deploy-preview-122--staging-tansu.netlify.app/project/?name=stellarpga

I still see some markdown marks like `##` for instance.

Oluwatos94:
> Can you double check @Oluwatos94 here I don't see the readme being rendered https://deploy-preview-122--staging-tansu.netlify.app/project/?name=stellarpga
> 
> I still see some markdown marks like `##` for instance.

Thanks! I would fix that now. That was an oversight.

tupui (approved):
Great that works! Will fix the linter on my end.

Shadow-MMN:
> Great that works! Will fix the linter on my end.

was still reviewing this 🥲

tupui:
Thanks @Shadow-MMN sorry for that, we can do a follow up if there was anything needed to be changed. I am making a presentation now and sort of needed to move fast.

PR #123, opened 2026-04-29
fix: show key file processing state

### LINKED ISSUE #106: Loading indicator on anonymous vote decoding

ComputerOracle:
Adds a loading state while processing the uploaded key file.

When decoding large ballots, there can be a short delay and the UI previously appeared unresponsive. This change shows a loading indicator during that time so it’s clear that processing is in progress.

Once decoding finishes, the flow continues as before.

closes #106


tupui (approved):
Thanks, this works for me. Not the rotating animation but what you did in the end is also good for that.

PR #124, opened 2026-04-29
fix: highlight anonymous votes over max weight

### LINKED ISSUE #108: Anonymous votes inconsistency flagging

ComputerOracle:
Highlights decoded anonymous vote rows when the decoded vote weight is greater than the voter’s max allowed weight.

This is a visual-only indicator to help identify inconsistent anonymous vote data during review. It does not block the flow, change validation, or affect vote execution.

Closes #108


tupui (approved):
This is good thanks!

PR #139, opened 2026-05-10
ci: upload SBOMs with Filebase

### LINKED ISSUE #129: Remove Storacha from SBOM job
Missed that from the migration to filebase. I can add the token to the env as requested.

dipespandey:
## Summary
- replace the Storacha SBOM upload action with a Filebase IPFS RPC upload
- switch the SBOM summary output to the Filebase CID and gateway URL
- keep the generated SPDX and CycloneDX artifact upload unchanged

Closes #129

## Validation
- ruby -e "require 'yaml'; YAML.load_file('.github/workflows/sbom.yml'); puts 'yaml ok'"
- git diff --check
- checked that the SBOM workflow no longer references Storacha

### DIFF
--- a/.github/workflows/sbom.yml
+++ b/.github/workflows/sbom.yml
@@ -72,12 +72,9 @@ jobs:
           )"
           printf '%s\n' "${FILEBASE_RESPONSE}"
 
-          cid="$(printf '%s\n' "${FILEBASE_RESPONSE}" | jq -r 'select(.Hash != null) | select((.Name // "") == "") | .Hash' | tail -n 1)"
-          if [ -z "${cid}" ] || [ "${cid}" = "null" ]; then
-            cid="$(printf '%s\n' "${FILEBASE_RESPONSE}" | jq -r '.Hash // .Cid // empty' | tail -n 1)"
-          fi
+          cid="$(printf '%s\n' "${FILEBASE_RESPONSE}" | jq -r '.Hash // .Cid // empty' | tail -n 1)"
 
-          if [ -z "${cid}" ] || [ "${cid}" = "null" ]; then
+          if [ -z "${cid}" ]; then
             echo "Unable to determine Filebase directory CID from upload response" >&2
             exit 1
           fi
--- a/.github/workflows/sbom.yml
+++ b/.github/workflows/sbom.yml
@@ -62,24 +62,19 @@ jobs:
         env:
           FILEBASE_TOKEN: ${{ secrets.FILEBASE_TOKEN }}
         run: |
-          if [ -z "${FILEBASE_TOKEN}" ]; then
-            echo "FILEBASE_TOKEN secret is required to upload SBOMs to Filebase" >&2
-            exit 1
-          fi
+          FILEBASE_RESPONSE="$(
+            curl --silent --show-error --fail \
+              -X POST \
+              -H "Authorization: Bearer ${FILEBASE_TOKEN}" \
+              "https://rpc.filebase.io/api/v0/add?wrap-with-directory=true" \
+              -F "file=@sbom-release/sbom.spdx.json;filename=sbom.spdx.json;type=application/json" \
+              -F "file=@sbom-release/sbom.cyclonedx.json;filename=sbom.cyclonedx.json;type=application/json"
+          )"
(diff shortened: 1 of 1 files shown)


.github/workflows/sbom.yml
```diff
@@ -57,22 +57,49 @@ jobs:
...
-          proof: ${{ secrets.STORACHA_PROOF }}
+      - name: Upload SBOMs to Filebase
+        id: filebase
+        env:
+          FILEBASE_TOKEN: ${{ secrets.FILEBASE_TOKEN }}
+        run: |
+          if [ -z "${FILEBASE_TOKEN}" ]; then
+            echo "FILEBASE_TOKEN secret is required to upload SBOMs to Filebase" >&2
+            exit 1
+          fi
```
tupui: Not needed

.github/workflows/sbom.yml
```diff
@@ -57,22 +57,49 @@ jobs:
...
+
+          response_file="$(mktemp)"
```
tupui: No don't use a file for this. Look I have this elsewhere and it's simpler

```bash
FILEBASE_RESPONSE="$(
  curl --silent --show-error --fail \
    -X POST \
    -H "Authorization: Bearer ${FILEBASE_IPFS_API_KEY}" \
    "https://rpc.filebase.io/api/v0/add?wrap-with-directory=true" \
    -F "file=@${PROPOSAL_FILE};filename=proposal.md;type=text/markdown" \
    -F "file=@${OUTCOMES_FILE};filename=outcomes.json;type=application/json" \
    -F "file=@${DISCUSSION_FILE};filename=proposal_discussion.md;type=text/markdown"
)"

CID="$(printf '%s\n' "${FILEBASE_RESPONSE}" | jq -r '.Hash // .Cid // empty' | tail -n 1)"
```

https://github.com/SCF-Public-Goods-Maintenance/scf-public-goods-maintenance.github.io/blob/main/.github/workflows/pg-award-proposal-onchain.yml#L137-L147

.github/workflows/sbom.yml
```diff
@@ -57,22 +57,49 @@ jobs:
...
+          curl --silent --show-error --fail-with-body \
+            -X POST \
+            -H "Authorization: Bearer ${FILEBASE_TOKEN}" \
+            "https://rpc.filebase.io/api/v0/add?wrap-with-directory=true&cid-version=1" \
```
tupui: Don't use v1. See my query above.

tupui (changes requested):
Thanks, some minor things

dipespandey:
Updated per review in dcf97ba:

- removed the explicit missing-token guard
- switched the Filebase response handling to command substitution instead of a temp file
- removed cid-version=1 from the Filebase add endpoint

I kept the existing FILEBASE_TOKEN secret name since this repo already uses that naming elsewhere.

.github/workflows/sbom.yml
```diff
@@ -57,22 +57,44 @@ jobs:
...
+
+          cid="$(printf '%s\n' "${FILEBASE_RESPONSE}" | jq -r 'select(.Hash != null) | select((.Name // "") == "") | .Hash' | tail -n 1)"
+          if [ -z "${cid}" ] || [ "${cid}" = "null" ]; then
+            cid="$(printf '%s\n' "${FILEBASE_RESPONSE}" | jq -r '.Hash // .Cid // empty' | tail -n 1)"
+          fi
+
+          if [ -z "${cid}" ] || [ "${cid}" = "null" ]; then
+            echo "Unable to determine Filebase directory CID from upload response" >&2
+            exit 1
+          fi
```
tupui: Does not make much sense to have these 2 if like that. Looks like some leftovers.

tupui (changes requested):

dipespandey:
Updated this as requested in 645c1c3. The CID extraction now uses the single Filebase response parsing pattern from your example, with the duplicate fallback removed.

tupui (approved):
Thanks, looks good now 👍

tupui:
If you are interested, same thing to do on the dapp-ipfs workflow @dipespandey 😃

dipespandey:
Thanks, I opened the follow-up for the `dapp-ipfs` workflow here: #141

PR #140, opened 2026-05-12
Remove vote

Oluwatos94:
Summary of what changed:

  - VoterInfo.tsx:  updated to CoI modal style: bg-[#F8F6FB] row, font-mono text-sm text-primary address, accepts
  optional action slot for the Remove button
  - RemoveVoteModal.tsx: uses <VoterInfo action={...} /> with the red Remove button passed as the action
  - ConflictOfInterestModal.tsx: renderRow now uses <VoterInfo /> too, so both modals share the same row style


tupui (changes requested):
You need to remove the changes to the bindings and the contract. These are up to date and now you add twice the function if you look carefully. All is there already.

tupui (approved):

PR #141, opened 2026-05-12
ci: upload dApp to IPFS with Filebase

dipespandey:
## Summary
- replace the dApp IPFS Storacha upload action with a Filebase RPC upload
- preserve `dist` relative paths in the multipart upload so nested assets keep their gateway paths
- update the workflow summary to report the Filebase CID and gateway URL

Follow-up to #139 and the maintainer note about applying the same Filebase migration to the `dapp-ipfs` workflow.

## Validation
- `ruby -e "require 'yaml'; YAML.load_file('.github/workflows/dapp-ipfs.yml'); puts 'yaml ok'"`
- `ruby -ryaml -e 'steps = YAML.load_file(".github/workflows/dapp-ipfs.yml").dig("jobs", "ipds", "steps"); puts steps.find { |step| step["name"] == "Upload dApp to Filebase" }["run"]' | bash -n`
- `git diff --check`
- confirmed `.github/workflows/dapp-ipfs.yml` no longer references Storacha
- ran the Filebase upload step locally with a mocked `curl` response to verify relative filenames, CID extraction, and GitHub output generation

### DIFF
--- a/.github/workflows/dapp-ipfs.yml
+++ b/.github/workflows/dapp-ipfs.yml
@@ -58,19 +58,61 @@ jobs:
           path: |
             ./dapp/dist/*
 
-      - uses: storacha/add-to-web3@892505d8e70c79336721485e5500155c17a728e0
-        id: storacha
-        with:
-          path_to_add: "./dapp/dist"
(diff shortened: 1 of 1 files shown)


tupui (approved):
Looks good, let's try it out

PR #143, opened 2026-05-15
Fix token based voting UI and add scaling to units

tupui:
Some UI fixes and minor scaling adjustments on the contract.

*Also removes the migration for good measures*


Shadow-MMN:
Implementation looks solid and solves the issue. Only minor note, the UI still says 10 XLM collateral for badge voting while the contract uses 2 XLM. Cosmetic only, not blocking.

Looks safe to merge from my review (just the Playwright test though).

tupui:
Thanks I will fix this 👍

PR #145, opened 2026-05-21
ci: add trivy filesystem jobs

### LINKED ISSUE #127: Add vulnerability scanning on dependencies
On the existing SBOM action, we can run this

https://github.com/anchore/grype
https://oss.anchore.com/docs/guides/vulnerability/getting-started/

The upload to IPFS should be just for tags otherwise we need to run periodically, once a week to start.

Also upload to IPFS to change to use filebase. (Ping me for the token to be present on the env.)

prempador:
## Summary
  Adds a dedicated **Trivy filesystem scan** workflow that runs on pull requests, pushes to `main`, and version tags (`v*`). Vulnerability scanning is separated from SBOM generation so each job has a single responsibility.
  The new scan uses [`aquasecurity/trivy-action`](https://github.com/aquasecurity/trivy-action) pinned to commit `ed142fd0673e97e23eac54620cfb913e5ce36c25` (v0.36.0), with `actions/checkout` pinned the same way as other release workflows. That gives us built-in vulnerability DB caching (important for PR feedback) and clearer CI gating via action inputs, without relying on mutable version tags. The SBOM workflow keeps its existing manual Trivy install, since it only runs on release tags and has a custom artifact/IPFS pipeline.
**Behavior change:** the old SBOM job ran `trivy fs .` with default settings (exit code 0, so it never failed). The new workflow fails on **CRITICAL** and **HIGH** findings with no available fix  (`ignore-unfixed: true`).
  ## Changes
  - **Add** `.github/workflows/trivy-fs.yml` — standalone filesystem vulnerability scan
  - **Update** `.github/workflows/sbom.yml` — remove the redundant `Scanning` step; SBOM generation unchanged
  ## Test plan
  - [ ] Open this PR and confirm the **Trivy filesystem scan** workflow runs and completes
  - [ ] Merge to `main` and confirm the workflow runs on the default-branch push
  - [ ] Tag a release (`v*`) and confirm both **Trivy filesystem scan** and **Generate SPDX and CycloneDX SBOM** run independently
  - [ ] Confirm the SBOM workflow still produces SPDX/CycloneDX artifacts and uploads to Filebase/IPFS as before

 ## Upgrading vulnerable packages that would block this PR: 

  bun update alone wasn’t enough — @stellar/stellar-sdk pins axios@1.15.0 and Astro pulls devalue@5.8.0 transitively. 
  Also bumped root @stellar/stellar-sdk to ^15.1.0.

Closes #127

### DIFF
--- a/.github/workflows/sbom.yml
+++ b/.github/workflows/sbom.yml
@@ -37,10 +37,6 @@ jobs:
         run: |
           wget -qO - https://raw.githubusercontent.com/aquasecurity/trivy/main/contrib/install.sh | sh -s -- -b /usr/local/bin
 
-      - name: Scanning
-        run: |
-          trivy fs .
-
       - name: Generate SBOMs
         run: |
           trivy fs --format spdx-json --output sbom.spdx.json .
--- a/.github/workflows/trivy-fs.yml
+++ b/.github/workflows/trivy-fs.yml
@@ -1,0 +1,35 @@
+name: Trivy filesystem scan
+
+permissions:
+  contents: read
+
+on:
+  pull_request:
+  push:
+    branches: [ main ]
+    tags:
+      - 'v*'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  trivy-fs:
+    name: Trivy filesystem scan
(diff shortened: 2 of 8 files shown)


tupui (approved):
Thanks 😊 failing tests are not linked no worries

PR #149, opened 2026-05-27
feat(governance): show "Already Voted" button state for voted proposals

### LINKED ISSUE #87: Already voted indication
On the proposals list and on a given proposal, we should adjust the button saying VOTE to be Already voted or something in case the connected user has already cast their vote.

e.g. https://testnet.tansu.dev/governance/?name=stellarpga this button:

Image
Image

Shadow-MMN:
## Summary

Fixes the UX issue where the "VOTE" button does not reflect whether the connected user has already voted on a proposal.

## Changes

- Added `hasUserVoted` helper in `utils/utils.ts` to derive voted state from proposal's `voteStatus.voters` arrays
- Updated vote button in both `ProposalCard.tsx` (list view) and `ProposalTitle.tsx` (detail view) to:
  - Show "Already Voted" label when user has voted
  - Disable the button in that state
  - Use secondary button styling for the voted state
  - Hide the vote icon for better readability on secondary button
- Removed redundant `useState` for `isVoted` in `ProposalPage.tsx` - value now derived directly from proposal data
- Updated `VotingModal.tsx` to use `onVoteSuccess` callback instead of `setIsVoted` for cleaner data refresh
- Updated `ProposalList.tsx` with `onVoteSuccess` callback to refresh proposal list after voting

## Before voting : 

Screenshot 2026-05-27 at 01 35 51

## After voting : 

Screenshot 2026-05-27 at 01 36 40

The `Already voted` also shows for the maintainer who proposed the proposal. 
closes #87

### DIFF
--- a/dapp/packages/tansu/tsconfig.json
+++ b/dapp/packages/tansu/tsconfig.json
@@ -23,7 +23,7 @@
     // "moduleDetection": "auto",                        /* Control what method is used to detect module-format JS files. */
     /* Modules */
     "module": "NodeNext" /* Specify what module code is generated. */,
-    // "rootDir": "./",                                  /* Specify the root folder within your source files. */
+    "rootDir": "./src" /* Specify the root folder within your source files. */,
     "moduleResolution": "nodenext" /* Specify how TypeScript looks up a file from a given module specifier. */,
     // "baseUrl": "./",                                  /* Specify the base directory to resolve non-relative module names. */
     // "paths": {},                                      /* Specify a set of entries that re-map imports to additional lookup locations. */
--- a/dapp/packages/tansu/tsconfig.json
+++ b/dapp/packages/tansu/tsconfig.json
@@ -23,7 +23,7 @@
     // "moduleDetection": "auto",                        /* Control what method is used to detect module-format JS files. */
     /* Modules */
     "module": "NodeNext" /* Specify what module code is generated. */,
-    "rootDir": "./src" /* Specify the root folder within your source files. */,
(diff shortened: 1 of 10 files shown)


dapp/packages/tansu/tsconfig.json
```diff
@@ -23,7 +23,7 @@
     // "moduleDetection": "auto",                        /* Control what method is used to detect module-format JS files. */
     /* Modules */
     "module": "NodeNext" /* Specify what module code is generated. */,
-    // "rootDir": "./",                                  /* Specify the root folder within your source files. */
+    "rootDir": "./src" /* Specify the root folder within your source files. */,
```
tupui: Why change this?

Shadow-MMN: I changed it because TypeScript was complaining about files being outside the inferred root during compilation, and Copilot suggested setting `"rootDir": "./src"` as a quick fix.

Looking at it again, I realize this changes the compiler boundary and output structure rather than fixing the actual underlying issue. I should probably have resolved the specific import/include problem instead of tightening `rootDir`.

I can revert this change and fix the original typing issue in a more targeted way if you prefer.

tupui: Yes please. This way we avoid breaking anything else

Shadow-MMN:
## Already voted button on the proposal list :
Screenshot 2026-05-27 at 10 45 36

Shadow-MMN:
## Reduced size :
Screenshot 2026-05-27 at 11 00 59

tupui (approved):
Thanks this is good 👍

PR #150, opened 2026-05-27
feat: show all outcome weights and seeds in anonymous voting tallies

### LINKED ISSUE #137: Full voter information in anonymous voting tallies
At the moment we only show a single weight and seed. This is ok when people don't cheat. But we actually want to show all outcome weights and seeds. It should then be clear that only a single outcome has a weight different than 0 and this is how we get the outcome selected by a vote. If there are multiple weights different than 0, then we don't know the outcome and have to mark the specific vote as problematic.

We already have some logic to flag a vote which has a power greater than the allowed voting power. So use the same highlighting.

JinadJay:
- Add outcomeWeights and outcomeSeeds arrays to DecodedVote interface
- Capture all outcome weights and seeds for each voter in computeAnonymousVotingData
- Display all outcome weights and seeds in AnonymousTalliesDisplay table
- Highlight votes with multiple non-zero weights as problematic (yellow background)
- Update CSV export to include all outcome weights and seeds

Fixes #137

### DIFF
--- a/dapp/src/components/page/proposal/AnonymousTalliesDisplay.tsx
+++ b/dapp/src/components/page/proposal/AnonymousTalliesDisplay.tsx
@@ -84,9 +84,9 @@ const AnonymousTalliesDisplay: React.FC<Props> = ({
                   <tr className="bg-zinc-100 text-left">
                     <th className="p-2">Address</th>
                     <th>Vote</th>
-                    <th>Weight</th>
+                    <th>Weights (A/R/Abs)</th>
                     <th>Max</th>
-                    <th>Seed</th>
+                    <th>Seeds (A/R/Abs)</th>
                   </tr>
                 </thead>
                 <tbody>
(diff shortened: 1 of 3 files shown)


JinadJay:
#Consulting-Manao

tupui (commented):
Thanks @JinadJay from a cursory look it seems good. @Shadow-MMN if you could have a look at it as you know that part better in the frontend.

Shadow-MMN:
> Thanks @JinadJay from a cursory look it seems good. @Shadow-MMN if you could have a look at it as you know that part better in the frontend.

I will have a look at it now.

Shadow-MMN:
@tupui , looks good to me also. I tested the it locally and it works good. Couldn't manually test this `Highlight votes with multiple non-zero weights as problematic (yellow background)` cause it is kinda hard or impossible to vote twice 😅. Safe to merge no breaking of the codebase.

Shadow-MMN (approved):
Works well after testing. Well done @JinadJay and happy coding. 😊

tupui (approved):

tupui:
Thank you @JinadJay and @Shadow-MMN for the review

PR #151, opened 2026-05-27
feat: support batching conflict of interest addresses

### LINKED ISSUE #130: Add/remove multiple addresses at once in conflict of interest list
Follow up of #94

We should be able to add or remove multiple addresses at the same time from the UI.

Shadow-MMN:
## Summary

Improves the Conflict of Interest address management flow by allowing users to batch multiple addresses before execution/signing.

Previously, clicking “Add” immediately triggered wallet signing for a single address. This PR changes the flow so users can queue multiple addresses locally, review them, and execute them together in one action.

## Changes

* Added pending address queue state
* Updated “Add” behavior to append addresses instead of executing immediately
* Added pending address list UI
* Added remove action for queued addresses
* Added validation for:

  * duplicate addresses
  * invalid addresses
  * already existing addresses
* Added batch execution flow via “Apply Changes”
* Added loading/disabled states during execution
* Cleared pending queue after successful execution
* Preserved existing single/bulk remove functionality

## UX Improvements

* Reduces repeated wallet signing prompts
* Makes multi-address management significantly faster
* Gives users a review step before execution

## Notes

* Existing remove functionality remains unchanged
* Existing bulk remove implementation remains unchanged
* Follows existing project patterns/components where possible

## Add button now adds to an address to a pending list so that user can add more and sign once :
Screenshot 2026-05-27 at 11 55 48

## Current list can be selected via checkbox and removed once by the user : 
Screenshot 2026-05-27 at 11 41 44

Closes #130

### DIFF
--- a/dapp/package.json
+++ b/dapp/package.json
@@ -38,7 +38,6 @@
     "react-dom": "^19.2.6",
     "react18-json-view": "0.2.10",
     "recharts": "3.8.1",
-    "tmp": "^0.2.6",
     "toml": "^4.1.1",
     "typescript": "^6.0.3"
   },
--- a/dapp/package.json
+++ b/dapp/package.json
@@ -38,6 +38,7 @@
     "react-dom": "^19.2.6",
     "react18-json-view": "0.2.10",
     "recharts": "3.8.1",
+    "tmp": "^0.2.6",
     "toml": "^4.1.1",
     "typescript": "^6.0.3"
   },
--- a/dapp/bun.lock
+++ b/dapp/bun.lock
@@ -43,6 +43,7 @@
         "react-dom": "^19.2.6",
         "react18-json-view": "0.2.10",
         "recharts": "3.8.1",
+        "tmp": "^0.2.6",
         "toml": "^4.1.1",
         "typescript": "^6.0.3",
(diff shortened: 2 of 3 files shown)


dapp/package.json
```diff
@@ -38,6 +38,7 @@
     "react-dom": "^19.2.6",
     "react18-json-view": "0.2.10",
     "recharts": "3.8.1",
+    "tmp": "^0.2.6",
```
tupui: Just to be sure, do we really need this? In general we should try to limit the number of our dependencies

Shadow-MMN: I added `tmp` because the Trivy scan was failing due to a vulnerable transitive dependency resolution. Adding `tmp` caused Bun to resolve and dedupe the dependency tree to a patched version, which made the scan pass. I added it not for runtime functionality in the app itself, but to address the dependency vulnerability detected during CI.

Shadow-MMN:
@tupui , I removed the tmp. Yh it was unnecessary. Thank you for the review. Next time I will be very particulate about adding dependencies.

tupui (approved):
This looks great now thanks 🙏

PR #154, opened 2026-05-28
fix: rename `Contract` contracttype to `ContractRef` to avoid `contractimport!` collision

### LINKED ISSUE #152: `Contract` contracttype name collides with generated `Contract` trait in `contractimport!`
## Problem

The Tansu wasm spec contains two items named `Contract`:

1. `#[contract] pub struct Contract` — the contract itself, from `contracts/tansu/src/contract_tansu.rs`. This becomes a generated `pub trait Contract` (matching the interface) in every consumer that does `soroban_sdk::contractimport!(file = "tansu.wasm")`.
2. `#[contracttype(export = false)] pub struct Contract` — a typed-`Address` wrapper used as the parameter type for `set_nqg_contract`, `set_domain_contract`, `set_collateral_contract`.

Both end up in the same module when a downstream contract imports Tansu's wasm, producing:

```
error[E0428]: the name `Contract` is defined multiple times
   |
   = note: `Contract` must be defined only once in the type namespace of this module
```

This blocks *any* external Soroban contract from using `contractimport!` on a Tansu release wasm — i.e. from getting typed `Client`, `Proposal`, `OutcomeContract`, `ProposalStatus` etc. bindings without hand-copying the Tansu types.

## Repro

In a new Soroban contract crate, depending on `soroban-sdk = "25.3.1"`:

```rust
soroban_sdk::contractimport!(file = "tansu_v2.0.2.wasm");
```

…with `tansu_v2.0.2.wasm` taken from the `v2.0.2_contracts_tansu_tansu_cli25.1.0` release asset. `cargo check` fails with the error above plus cascading `E0574`/`E0782`.

The same failure occurs with `soroban-sdk-tools::contractimport!` (which wraps the standard one).

The `experimental_spec_shaking_v2` feature on `soroban-sdk` does not help: the `Contract` contracttype is reachable from the trait (it's a function parameter type), so reachability shaking keeps it.

## Why it matters

Downstream contracts that need to read Tansu proposals (e.g. a DAO-gated registry manager) currently have to either:

- hand-mirror `Proposal` / `OutcomeContract` / `ProposalStatus` / `VoteData` types verbatim and call `get_proposal` via untyped `env.invoke_contract`, with the well-known risk of silent drift as Tansu evolves, **or**
- maintain a side-channel "stub" wasm whose sole job is to expose those types for `contractimport!`.

Both add boilerplate Tansu owns nothing of. The first one is dangerous (positional encoding will silently break on field reorder); the second one is fine but means every consumer reimplements it.

## Proposed fix

Rename one of the two `Contract` items. Either of the following is sufficient — both are non-breaking at the wire level (struct/enum tags are positional, the *name* doesn't go on the wire), only ABI-breaking for code that uses the symbol by name.

**(a) Rename the contract struct.** In `contracts/tansu/src/lib.rs` (or wherever `#[contract] pub struct Contract` lives):

```rust
#[contract]
pub struct TansuDao;       // or `Tansu`
```

…and update the `impl` blocks accordingly. Downstream `contractimport!` then generates `pub trait TansuDao` instead of `pub trait Contract`, and the `Contract` contracttype struct stays.

**(b) Rename the contracttype wrapper.** In `contracts/tansu/src/types.rs`:

```rust
#[contracttype(export = false)]
pub struct ContractRef {   // was: pub struct Contract
    // ...
}
```

…and update the `set_*_contract` parameter types. Downstream gets a clean `pub trait Contract` + `pub struct ContractRef`.

(a) is the more conventional choice (a Soroban contract struct named after its purpose rather than the generic word `Contract`). I have a mild preference for (a) but happy with either — let me know which you'd take and I can send a PR.

## Context

Hit this while building `registry-tansu-manager` in `theahaco/scaffold-stellar` ([PR #518](https://github.com/theahaco/scaffold-stellar/pull/518)) — a Tansu-DAO-gated manager for the Stellar Registry that gates registry operations behind approved Tansu proposals. We worked around it by importing a local stub's wasm instead of real Tansu, but plan to switch over once a release with the rename lands.

willemneal:
## Summary

Fixes #152.

- Rename `types::Contract` → `types::ContractRef` (the `#[contracttype]` struct used as the parameter to `set_domain_contract` / `set_collateral_contract` / `set_nqg_contract`).
- Update all internal references (`contract_tansu.rs`, `lib.rs`, tests).
- Regenerate TypeScript bindings under `dapp/packages/tansu`.

## Why

`soroban-spec-rust` hardcodes `pub trait Contract` for the generated interface when downstream crates use `soroban_sdk::contractimport!(file = "tansu.wasm")` (see `soroban-spec-rust/src/lib.rs#L119` in the 26.x line). The Tansu spec also exports a `Contract` contracttype, so both symbols collide in the same module:

```
error[E0428]: the name `Contract` is defined multiple times
```

This blocks any external Soroban contract from importing the Tansu wasm to get typed `Client`, `Proposal`, `OutcomeContract`, `ProposalStatus`, `VoteData` bindings without hand-copying types.

## Compatibility

- **Wire format: unchanged.** Contracttype struct/enum tags are positional; the field name is not encoded on the wire. Existing on-chain state and existing clients calling `set_*_contract` continue to work.
- **Source-level ABI: breaking by name only.** Any Rust crate that referenced `tansu::Contract` (the contracttype) must update to `tansu::ContractRef`. Downstream consumers now get a clean `pub trait Contract` (the contract interface) plus `pub struct ContractRef` from `contractimport!`.

## Test plan

- [x] `cargo check --manifest-path contracts/tansu/Cargo.toml`
- [x] `cargo test --manifest-path contracts/tansu/Cargo.toml --release` — all 54 tests pass
- [x] `cargo build --manifest-path contracts/tansu/Cargo.toml --release --target wasm32v1-none`
- [x] Regenerate TypeScript bindings (`stellar contract bindings typescript`) — `Contract` interface becomes `ContractRef`; no dapp consumer references the old name directly.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


tupui (approved):
Oh nice catch! Thanks a lot (yes having some windows there and there 😆)

PR #155, opened 2026-05-28
feat: per-project governance overrides (min_voting_period, execute_delay)

willemneal:
## Summary

Per-project governance timing overrides:

- `min_voting_period: Option<u64>` — minimum seconds between proposal creation and `voting_ends_at`. Defaults to the global `MIN_VOTING_PERIOD` (24h).
- `execute_delay: Option<u64>` — seconds between `voting_ends_at` and when `execute()` becomes callable. Defaults to the global `TIMELOCK_DELAY` (24h).

Both can be set at registration and changed later through `update_config`. Together they let a project drive a proposal end-to-end in seconds instead of the ~25-hour floor. Use cases: testnet flows, internal-tool DAOs, fast-iteration governance experiments.

## API changes

```rust
fn register(
    env, maintainer, name, maintainers, url, ipfs,
    min_voting_period: Option<u64>,  // NEW — None = use the global default
    execute_delay: Option<u64>,      // NEW
) -> Bytes;

fn update_config(
    env, maintainer, key, maintainers, url, ipfs,
    min_voting_period: Option<u64>,  // NEW — None = leave current value unchanged
    execute_delay: Option<u64>,      // NEW
);
```

- Note the `None` asymmetry: at `register` it means "no override"; at `update_config` it means "don't touch" — a plain config update never alters governance. To restore a default, pass its value explicitly (e.g. `Some(86400)`).
- Overrides validate `Some(v)` in `1..=MAX_VOTING_PERIOD` (30 days). Out-of-range panics `ContractErrors::InvalidVotingPeriod = 214` (212/213 were taken by the evidence errors on main).
- **Breaking:** `update_config` gains two args (as `register` already did in this PR). All dapp call sites are updated; external integrators must regenerate bindings after the contract upgrade.
- No public getters (per review) — values are observable through the events below.

## Security model for updates

Governance updates via `update_config` are maintainer-gated, and two mechanisms keep them from being abused:

1. **Per-proposal timelock snapshot.** `create_proposal` snapshots the project's effective `execute_delay` into `ProjectKey::ProposalExecuteDelay(project_key, proposal_id)`; `execute()` reads the snapshot (falling back to `TIMELOCK_DELAY`). An update can never change the reaction window of an in-flight proposal.
2. **Notice window for loosening.** Tightening (every provided value ≥ its current effective one) applies immediately. Loosening is stored as `PendingGovernance` and only activates after current `min_voting_period + execute_delay` seconds — a rule change can never pass faster than a proposal under the rules it replaces. Pending updates apply lazily on the next `create_proposal`/`update_config`; no activation transaction needed.

Without these, a single maintainer could sandwich: drop both floors to 1s → propose → self-approve → execute arbitrary `outcome_contracts` in ~2 ledgers → restore.

## Events

- `ProjectGovernanceUpdated { project_key, maintainer, min_voting_period, execute_delay, activates_at }` — emitted whenever `update_config` receives governance values; `activates_at` = now for tightening, end of notice window for loosening. (`ProjectConfigUpdated` fires as before for the config part.)
- `ProposalCreated` now carries `execute_delay` — the snapshotted timelock — so voters see a proposal's real reaction window at vote time.

## Storage

- `ProjectKey::MinVotingPeriod(Bytes) -> u64`
- `ProjectKey::ExecuteDelay(Bytes) -> u64`
- `ProjectKey::ProposalExecuteDelay(Bytes, u32) -> u64` — snapshot at proposal creation
- `ProjectKey::PendingGovernance(Bytes) -> PendingGovernance { min_voting_period, execute_delay, activates_at }`

## Where the overrides take effect

- `create_proposal` and `update_config` promote any activated pending update first; `create_proposal` then reads `MinVotingPeriod` for the `voting_ends_at` lower bound and snapshots `ExecuteDelay`.
- `execute` gates on the proposal's snapshot.
- **Intentionally unchanged:** the admin-upgrade timelock in `propose_upgrade` still uses the hardcoded `TIMELOCK_DELAY`. Different threat model — admin compromise window — and not something a project's maintainers should be able to shorten.

## dapp

- `updateConfigFlow`/`createSignedUpdateConfigTransaction` pass the new params (`undefined` = leave unchanged); `createProjectFlow` already passed them for `register`. `UpdateConfigModal` unchanged — no UI form fields yet.
- On-chain activity decoding (`constants/onchain.ts`) lists the new arg names for `register`/`update_config`; positional decode in `OnChainActivityService` is unaffected (args append at the end).
- Bindings regenerated with stellar-cli 27. This also fixed the donation failures seen during review: the earlier regen pinned `@stellar/stellar-sdk@^14.5.0` while the dapp runs v16, mixing SDK versions across transaction assembly; the package now pins `^16.1.0`. Stale `dist/src/` duplicates removed.

## Test plan

- [x] `cargo test` — 91 pass, including:
  - register-time overrides: defaults / stored / rejects zero / rejects above max, floor applied in `create_proposal`, 60s delay gates `execute()` at exactly `voting_ends_at + 60`
  - `update_config` governance: tighten applies instantly, plain config update leaves overrides untouched, loosening goes pending and applies after the notice window, invalid values rejected, non-maintainer rejected
  - `execute_delay_snapshot_shields_in_flight_proposals` — delay change after creation doesn't move an open proposal's timelock
  - `execute_delay_loosening_waits_out_notice` — loosening inert until the notice window elapses; only proposals created afterwards get the new delay
- [x] `cargo clippy --all-targets` clean, `cargo fmt` clean
- [x] `stellar contract build --optimize`
- [x] Bindings regenerated + package builds (`tsc`); `bun scripts/validate-contract-errors.js` passes
- [x] dapp: vitest 248 pass (under `TZ=UTC`; 3 pre-existing TZ-sensitive date tests fail in non-UTC shells on main too), `astro build` completes
- CI: contract tests, Playwright E2E, pre-commit green. Trivy + eslint failures reproduce identically on main (dep CVEs / `typescript-estree` crash) — not from this branch.

## Out of scope

- Per-proposal `execute_delay` chosen at `create_proposal` (risk-proportional timelocks). The snapshot + project floor lays the groundwork; follow-up if wanted.
- dapp UI fields for the overrides (plumbing exists in `FlowService`).
- The admin-upgrade timelock stays hardcoded (see above).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01EJg2YNTauCJJ8Vj8nr19jr


dapp/packages/tansu/dist/index.d.ts
```diff
@@ -1189,12 +1197,16 @@ export interface Client {
       maintainers,
       url,
       ipfs,
+      min_voting_period,
+      execute_delay,
```
tupui, line 1255: @Shadow-MMN just checking with you if this breaks on the frontend side or not? With the option I would think not but just checking with you.

Shadow-MMN: let me check it now .

Shadow-MMN: I was able to create a proposal and even vote successfully, but couldn't make donations. 

#### These are the errors I got when I tried making a donation :
Screenshot 2026-05-28 at 21 01 41
Screenshot 2026-05-28 at 21 01 49

tupui: Donation should not be linked though no?

contracts/tansu/src/lib.rs
```diff
@@ -75,15 +80,22 @@ pub trait MembershipTrait {
...
         maintainer: Address,
         name: String,
         maintainers: Vec<Address>,
         url: String,
         ipfs: String,
+        min_voting_period: Option<u64>,
+        execute_delay: Option<u64>,
     ) -> Bytes;
 
+    fn get_min_voting_period(env: Env, project_key: Bytes) -> u64;
```
tupui: Do we really need these 2 in the public API? I don't have much simple accessor otherwise as you can see elsewhere. I tend to just inline storage calls.

contracts/tansu/src/errors.rs
```diff
@@ -26,6 +26,7 @@ pub enum ContractErrors {
     VoterWeight = 209,
     VoteLimitExceeded = 210,
     VoterConflicted = 211,
+    InvalidVotingPeriod = 212,
```
tupui: There is a mapping file to update. I know that thing is not ideal. (The linter failure)

dapp/packages/tansu/dist/index.d.ts
Shadow-MMN: I am not so sure but I switched back to main and tried the donation and it went through. So the failing must be coming from this pr.
Screenshot 2026-05-28 at 21 02 24

tupui:
@willemneal Ping in case you missed my review. Otherwise LGTM

tupui (approved):
@willemneal what is the status here? Now it looks good to me and happy to get in. Just noting that the update config function does not let people adjust the timings. Which can be ok.

tupui (approved):

tupui:
Seen offline, thanks 😊

PR #158, opened 2026-05-29
Add Radicle repository provider support

### LINKED ISSUE #18: Evaluate support for Radicle as another VCS

Septimus4:
Here I am again, it was less work than I thought.
But I never used radical so if there are more feature from it that I could/should integrate or support I am open to feedback

Add parsing, validation, and UI handling for Radicle RIDs and seed URLs.
Update TOML configuration, documentation, and tests to support Radicle maintainer aliases and repository workflows.

Closes #18


tupui (commented):
Thai is awesome 🤩 thanks a lot for working on it! Did you register a project to test?

@Shadow-MMN if you can review I would really appreciate. Thanks 🙏

Shadow-MMN:
> Thai is awesome 🤩 thanks a lot for working on it! Did you register a project to test?
> 
> @Shadow-MMN if you can review I would really appreciate. Thanks 🙏

On it

dapp/src/components/page/project/UpdateConfigModal.tsx
```diff
@@ -119,18 +122,34 @@ function mergeTomlData(
   existingDoc["ORG_URL"] = fields.orgUrl;
   existingDoc["ORG_LOGO"] = fields.orgLogo;
   existingDoc["ORG_DESCRIPTION"] = fields.orgDescription;
-  existingDoc["ORG_GITHUB"] = fields.isSoftwareProject
-    ? getRepositoryProjectPath(fields.githubRepoUrl)
-    : "";
+  if (fields.isSoftwareProject && fields.repositoryProvider === "radicle") {
```
Shadow-MMN, line 127: Saving an existing project config can unintentionally remove `ORG_REPOSITORY_SEED` because project.config.url is normalized to a `rad:RID` ( in `StateService.ts` around line 94). This causes the logic to fall into the deletion branch even for projects originally using a custom seed, breaking seed-based browsing and metadata fetches.

We should preserve the existing seed unless the user explicitly changes or removes it, rather than inferring removal from the normalized URL.

dapp/src/utils/editLinkFunctions.ts
```diff
@@ -158,13 +191,135 @@ function buildNormalizedRepositoryUrl(
...
+  try {
+    const parsedUrl = new URL(repoUrl);
+    if (
+      parsedUrl.protocol !== "https:" ||
+      (parsedUrl.port && parsedUrl.port !== "443")
+    ) {
+      return undefined;
+    }
+
+    const rid = extractRadicleRidFromUrl(parsedUrl);
```
Shadow-MMN: This is too permissive when parsing Radicle URLs. Any HTTPS URL with a `rad:` substring in the path or query, plus any dotted hostname, can be treated as a Radicle repo. That can make unrelated URLs pass validation and generate bogus Radicle config/links.

Can we tighten the parsing rules to only accept supported Radicle URL formats (e.g. `rad:RID` or known Radicle seed URL patterns) rather than matching arbitrary URLs that happen to contain rad: somewhere in the URL?

Shadow-MMN (changes requested):
Hey @Septimus4 , so sorry my review is coming late. I have checked the changes and I commented on my findings. I am having a bit of a challenge testing locally. Can we speak more on discord or telegram username - Shadow_MMN

dapp/src/utils/editLinkFunctions.ts
Septimus4: I will address the changes this evening. Yes we can have a discord call, what is your handle ?

Shadow-MMN: > I will address the changes this evening. Yes we can have a discord call, what is your handle ?

cool, here is my handle - shadow_mmn

Septimus4:
image

dapp/src/components/page/project/UpdateConfigModal.tsx
Shadow-MMN, line 127: The current logic can preserve a stale ORG_REPOSITORY_SEED when the repository changes.

If a user edits a project to point to a different Radicle RID or pastes a direct rad:... reference, and the new URL does not explicitly include a seed host, the existing ORG_REPOSITORY_SEED is carried forward unchanged. This can leave the project associated with a seed from a previous repository and cause browsing, metadata, or README fetches to target the wrong seed.

Can we ensure the seed is derived from the newly parsed URL when available, and otherwise clear or explicitly update the value rather than preserving the previous seed across repository changes?

Shadow-MMN (changes requested):
We are getting there. Great Job so far.

dapp/src/components/page/project/UpdateConfigModal.tsx
Septimus4: I am not 100% I understood correctly, is that okay, or you had something else in mind ?
If the new Radicle URL includes a seed host, ORG_REPOSITORY_SEED is set from that new URL.
If the URL has no seed and the RID changed, the old seed is deleted.
If the user changes the field from a seed URL to a direct rad: reference, the old seed is deleted.
If the form only has the already-normalized same rad:RID and was not changed, the existing seed is preserved.

Septimus4: I pushed the change, if it's okay I will squash

Septimus4:
locally the playwright tests succeed but seems to always timeout on the CI, are they known to do that ? should we investigate ?

tupui:
No worries there. e2e tests are broken in the CI. As long as you're both happy we can go ahead.

Shadow-MMN (commented):
Thanks for the update. I think this addresses the case where a different RID is selected and prevents a stale seed from being carried over.

There is still one edge case though: the current logic appears to key off whether the repository field value changed rather than whether the parsed RID changed. As a result, a user could switch between different representations of the same repository (for example, from a browse URL to a direct `rad:...` reference with the same RID), and `ORG_REPOSITORY_SEED` would be removed even though the repository itself has not changed.

In that scenario, the existing seed host is still valid and useful, so dropping it seems unnecessary. Would it make more sense to base this decision on RID changes rather than raw URL string changes, preserving the seed when the parsed RID remains the same?

tupui:
Let me know when this is ready to get in

Shadow-MMN (approved):
LGTM. Great work @Septimus4 . Checked the current changes and you have ratified all issues I found. As for the QA, I was able to test on my end and everything works well. Once again nice work and happy coding 😊.
@tupui we are safe to merge but feel free to add your own review.

Shadow-MMN:
> Let me know when this is ready to get in

you can now 😅

tupui (approved):
Getting this in and will try to test on my end. Thanks a lot for working on this! Fire we did this 🤩☄️

tupui:
@Septimus4 I could create a project but then I am seeing some errors in the console when pulling the commit history https://testnet.tansu.dev/project/?name=tansunew

Septimus4:
Can you share the errors ? I am not on my computer but from my phone I only see a CORS related error

tupui:
Nothing urgent also not anymore in front of my laptop. It was a 404

Septimus4:
Okay I will look into it this evening

Septimus4:
@tupui I think this is more likely a Radicle seed/setup issue for tansunew than an app issue. 
The error suggests the seed does not have the repo for that RID in its local Radicle storage. 
What we could improv eon the app side is to validate the seed/RID when saving the project, and show a clean “repo not available on configured Radicle seed” error instead of leaking the /var/lib/radicle/storage/... error.

Septimus4:
I can't check the config of tansunew but make sure Radicle RID + ORG_REPOSITORY_SEED is valid

tupui:
What do you mean? The RAD I registered exists

Septimus4:
I checked the RID against a few Radicle nodes.

This one fails:
https://radicle.network/nodes/seed.radicle.dev/rad%3AzssaAF91kxuquZmZCV2SiK2FNX6s → 404

These ones work:
https://radicle.network/nodes/iris.radicle.network/rad%3AzssaAF91kxuquZmZCV2SiK2FNX6s
https://radicle.network/nodes/rosa.radicle.network/rad%3AzssaAF91kxuquZmZCV2SiK2FNX6s

So the RAD exists, but not on every seed. That means the issue depends on what was configured in the project TOML.
I am not sure if a ORG_REPOSITORY_SEED fallback value is set in the TOML.

For this project, the Radicle URL should use one of the working nodes, for example:
https://radicle.network/nodes/iris.radicle.network/rad:zssaAF91kxuquZmZCV2SiK2FNX6s
or:
https://radicle.network/nodes/rosa.radicle.network/rad:zssaAF91kxuquZmZCV2SiK2FNX6s

but not just rad:zssaAF91kxuquZmZCV2SiK2FNX6s

is that clearer ?

tupui:
Ok yes but then we need to be clear about that for users. iris or rosa are ok as defaults. Otherwise we need to ask for the node to use. E.g. https://radicle.network/nodes/consulting-manao.radicle.garden Now Radicle proposes a node service: Garden.

Septimus4:
Yes, agreed.

The frontend asks for one repository URL, and for Radicle we parse that URL to extract both the RID and, when present, the seed/node host.
That made sense to me coming from the GitHub/GitLab model. With Git providers, the repo URL already contains the provider/host: `github.com/org/repo`, `gitlab.com/org/repo`, etc. So I treated Radicle similarly: if the user pastes a full Radicle explorer/node URL, we can infer both the repo and the node from one value.

So yes, we should make this clearer for users, I see two options:
1. Small frontend change: keep the single repository URL field, but improve the placeholder/tooltip for Radicle. Something like: “Paste the full Radicle node URL when possible, e.g. `https://radicle.network/nodes/iris.radicle.network/rad:...`. If only `rad:...` is provided, Tansu will use the default seed.”
2. Bigger UX change: split Radicle into two explicit fields: Radicle RID nad Radicle node/seed

tupui:
Yeah let's just do the simple thing for now with a better tooltip 👍 And we can change the default node to be iris. It would cover most use cases I think because this is a default seeding node with an open policy.

Septimus4:
Okay I will open PR tomorrow

Issue #112, opened 2026-04-28
Consider a Merkle distribution


tupui:
To repay collateral, we are limited with the number of `transfer`s we can do in a single smart contract execution. (Around 80 maybe, mostly due to the limit on events and write entries.)

Instead, we could switch to a claim mechanism where we let participants claim their collateral back.

We would then need to add a button after voting to let user claim their collateral. Or have a central claim button on the membership space might be more convenient. To discuss.

We could also consider re-thinking the collateral system and go back to using the memberships. We could set a different membership model as well. E.g. you put a collateral to be a member of a community (and can only leave and reclaim the collateral if you do not have pending votes maybe. Otherwise someone can become a member just to cast a vote.)

Septimus4:
@tupui Can I start looking into this?

One clarification: when you say community, do you mean a project, or a higher level entity that can group multiple projects?

From the current model membership/badges are project scoped and we do not have a separate community representation yet. So if you mean the second case, this may require a broader model change than just the collateral claim mechanism.

tupui:
I think I still need a bit of time to think about this one. I am going to revisit this around July I think as I am away in June. Same with #111 as both are linked.

If you have some thoughts do write them in the meantime.

There is also maybe a future where we have Nouns DAO coming to Stellar and instead I might want to rely on people having a Noun or putting one in custody.

Septimus4:
I can do that, is there any other issue I could start working on  ?

tupui:
Otherwise the supply chain ones maybe. But here as well that's about making some design proposals first. Like what a release would look like onchain. Eg Radicle has a concept of attesting a commit which I would also like to bring here and would be especially relevant for releases.

Septimus4:
you mean issue 21 and 26 ? I can look into them

tupui:
#25 and #26 yep

tupui:
#15 also linked for the release part.

tupui:
Closing in favor of https://radicle.network/nodes/rosa.radicle.network/rad%3AzssaAF91kxuquZmZCV2SiK2FNX6s/issues/c6a71ed20bd6bfd9af5f34c838135919c21ac2f4

PR #161, opened 2026-05-30
perf: optimize vote storage lookup complexity

### LINKED ISSUE #111: Vote casting addresses logic
We can change the vote casting logic by not having a list of addresses who voted and instead check that a vote exists or not. This would remove the limits on the vector of votes increasing as O(n) to O(1).

ComputerOracle:
Summary

This PR optimizes vote tracking by replacing vector-based vote counting with direct storage lookups.

Changes

* Added a dedicated vote counter for tracking total votes.
* Removed the need to load the full voter list when checking vote limits.
* Kept duplicate-vote protection through direct vote existence checks.
* Updated vote counting logic to use constant-time storage access.

Benefits

* Improves scalability as the number of voters grows.
* Reduces storage reads and computation costs.
* Changes vote limit checks from **O(n)** to **O(1)**.
* Preserves existing voting behavior and security guarantees.

Files Modified

* `contracts/tansu/src/types.rs`
* `contracts/tansu/src/contract_dao.rs`

closes #111

### DIFF
--- a/contracts/tansu/src/contract_dao.rs
+++ b/contracts/tansu/src/contract_dao.rs
@@ -307,10 +307,16 @@ impl DaoTrait for Tansu {
             &types::ProjectKey::Vote(project_key.clone(), proposal_id, proposer.clone()),
             &vote_,
         );
-        let voters = vec![&env, proposer.clone()];
+        // Proposer's abstain occupies slot 0; VoteNextIndex advances to 1.
+        // VoteCount is NOT written here: the proposer's weight=0 abstain was never
+        // counted in the original VoteCount, preserving DoS-limit semantics.
         env.storage().persistent().set(
-            &types::ProjectKey::Voters(project_key.clone(), proposal_id),
-            &voters,
+            &types::ProjectKey::VoteIndex(project_key.clone(), proposal_id, 0),
+            &proposer,
+        );
+        env.storage().persistent().set(
+            &types::ProjectKey::VoteNextIndex(project_key.clone(), proposal_id),
+            &1u32,
         );
 
         let proposal_tallies = if public_voting {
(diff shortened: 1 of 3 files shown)


tupui (changes requested):
Thanks @ComputerOracle 

There is a misunderstanding here as this is only addressing partially the issue. The real fix is about not storing a vector of voters at all.

ComputerOracle:
Thanks for the feedback
I'll go over the codebase again and make the necessary changes.

tupui:
I am not sure we are going the right direction with this index system.

In the end we might just want to have a classical Merkle Tree and that would address #112 at the same time. That is out of scope for this work.

I am going to call it a day here and keep the list for now. Thanks for trying this out still.

ComputerOracle:
That's completely fair, and thanks for taking the time to review it. I
appreciate the feedback even though the approach didn't end up being the
direction the project wants to take.

It was genuinely nice working on this and thinking through the problem. I
learned quite a bit from it.

Also, this might not even be the right place to ask 😅, but I've been
building a Soroban project called PredictX on the side—a decentralized
prediction marketplace—and I'm hoping to eventually submit it as a
repository for Stellar Wave.

Since you've seen quite a few projects in the ecosystem, do you have any
advice on what tends to make a project stand out to reviewers or
maintainers? I'm trying to make sure I'm focusing on the right things
beyond just getting features working.

Either way, thanks again for the feedback and for giving the idea a shot.
Hopefully we'll get the chance to collaborate again on another issue
sometime.

tupui:
Pleasure.

For your question to be honestly most contributors don't really care about the project. I have 2-3 people genuinely interested but the rest are just "points farmers". But if your project is real, you might want to consider pushing it on the Stellar discord and even consider going for an insta award or build award depending on the status. Fee free to send me a DM on Discord if you want to share more about your project and need guidance.

ComputerOracle:
Okay then, would send a message on Discord asap.
Thank youu

PR #167, opened 2026-06-07
chore(radicle): update default seed host to iris.radicle.network

Septimus4:
Update default seed host to iris.radicle.network
Improve descriptions to guide users on Radicle url handling


tupui (approved):
That's great thanks

PR #176, opened 2026-06-14
fix: derive project key from raw name to match on-chain contract

### LINKED ISSUE #173: The bug: project-key derivation mismatch on registration
Project registration sends the raw-case name (Tansu.register({ name: projectName })) while deriveProjectKey lowercases before hashing, so any project name containing uppercase letters is registered under a key the rest of the dapp can never reproduce — making the project permanently invisible/unusable in the UI.

ex:
- https://github.com/Consulting-Manao/tansu/blob/7744362d74e3e78fabbc7e9c11600ba20604ab68/dapp/src/service/FlowService.ts#L332
const tx = await Tansu.register({
  maintainer: publicKey,
  name: projectName,        // ← line 334: raw, e.g. "MyProject" (NOT lowercased)
  maintainers,
  url: normalizedRepositoryUrl,
  ipfs: cid,
});

- https://github.com/Consulting-Manao/tansu/blob/7744362d74e3e78fabbc7e9c11600ba20604ab68/contracts/tansu/src/contract_versioning.rs#L53
let name_b = name.to_bytes();                 // "MyProject" — original case
for b in name_b.iter() {
    if !(b.is_ascii_lowercase() || b.is_ascii_uppercase() || b.is_ascii_digit()) {
        panic_with_error!(&env, &errors::ContractErrors::InvalidProjectName);  // uppercase IS allowed
    }
}
let key: Bytes = env.crypto().keccak256(&name_b).into();   // ← key = keccak256("MyProject")

- Everywhere else, the frontend lowercases before hashing
https://github.com/Consulting-Manao/tansu/blob/7744362d74e3e78fabbc7e9c11600ba20604ab68/dapp/src/utils/projectKey.ts#L10
export function deriveProjectKey(projectName: string): Buffer {
  return Buffer.from(
    keccak256.create().update(projectName.toLowerCase()).digest(),  // ← keccak256("myproject")
  );
}

So a project registered as MyProject is stored on-chain under the first key, but the dapp forever looks it up under the second → it can't find it.

How to fix:
The fix is one line in https://github.com/Consulting-Manao/tansu/blob/7744362d74e3e78fabbc7e9c11600ba20604ab68/dapp/src/service/FlowService.ts#L332  - lowercase the name so registration matches deriveProjectKey: name: projectName.toLowerCase()

real-venus:
The contract derives the project key from keccak256 of the raw (case-sensitive) name, but the dapp lowercased the name before hashing. Projects registered with uppercase letters were stored under a key the dapp could never reproduce, making them unreachable in the UI.

Remove .toLowerCase() in all three key-derivation sites so the dapp derives the same key the contract stores:
- utils/projectKey.ts (deriveProjectKey)
- service/StateService.ts (setProjectId)
- service/OnChainActivityService.ts (register cache key)

Safe for existing lowercase projects (hashing an already-lowercase name is unchanged); recovers uppercase-named projects that were orphaned.

### DIFF
--- a/dapp/src/service/OnChainActivityService.ts
+++ b/dapp/src/service/OnChainActivityService.ts
@@ -185,10 +185,10 @@ export async function fetchOnChainActions(
           case "register": {
             projectName = paramToString(args[1]); // name is arg1 (maintainer is arg0)
             if (projectName) {
-              // keccak key (derived) – used by some off-chain calls
-              const keyHex = normalizeHex(
-                keccak_256(projectName.toLowerCase()),
-              );
+              // keccak key (derived) – used by some off-chain calls.
+              // Hash the raw name to match the on-chain contract (case-sensitive),
+              // consistent with deriveProjectKey().
+              const keyHex = normalizeHex(keccak_256(projectName));
               PROJECT_CACHE.set(keyHex!, projectName);
               details.name = projectName;
 
--- a/dapp/src/service/StateService.ts
+++ b/dapp/src/service/StateService.ts
@@ -73,7 +73,7 @@ function setProjectId(project_name: string): void {
   projectState.project_name = project_name;
 
   projectState.project_id = Buffer.from(
-    keccak256.create().update(project_name.toLowerCase()).digest(),
+    keccak256.create().update(project_name).digest(),
   );
 
   if (typeof window !== "undefined") {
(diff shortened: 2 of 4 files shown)


real-venus:
@tupui 
I am now ready for review

Shadow-MMN (approved):
LGTM everything works well. @real-venus nice work on this. 
@tupui, we are safe to merge.

tupui (approved):
Thank you for working on this!

PR #186, opened 2026-06-23
Add evidence handling and dApp integration for external evidence

### LINKED ISSUE #15: Store SBOM
xref https://github.com/tupui/soroban-versioning/issues/175

### LINKED ISSUE #25: Store release attestations
Artifact Attestations should be stored in IPFS and we should store the CIR onchain. This way release artifacts can be verified in a decentralized fashion.

### LINKED ISSUE #26: Store CVEs
Maintainers need a way to run an analysis on their project (and its dependencies). The resulting CVEs would be stored onchain and attached to a specific commit hash. We need to think about historical data as right now we only store the last hash (since it's a merkle tree and the last hash is enough to validate the descendent).

Septimus4:
Closes #15 · Closes #25 · Closes #26

Lets maintainers attach external supply-chain evidence (SBOM, vulnerability scan, build attestation) to a specific project commit. Artifacts live off-chain on IPFS; the contract stores only the content identifier (CID) plus a ledger timestamp. The whole flow:
produce → pin to IPFS → record on-chain → read & verify

```
(project_key, commit_hash, kind) -> IPFS CID        kind ∈ {Sbom, Cve, Attestation}
```

### What's included

**Contract**
- `set_evidence` / `get_evidence` for generic per-commit evidence, maintainer-gated.
- **Append-only history**: each `set_evidence` adds an indexed entry instead of overwriting, with a per-`(project, commit, kind)` counter. Re-scans of the same commit (the common CVE case) accumulate a full on-chain timeline. `get_evidence` returns the latest; `get_evidence_count` + `get_evidence_at` expose the history.
- **Durability**: entries are bumped toward the maximum persistent-storage TTL on write, and `bump_evidence` lets anyone permissionlessly keep a historical entry alive. Archived entries remain restorable, so history is never lost.
- `EvidenceSet` event (carries the entry `version`) for on-chain observability.

**dApp**
- `EvidenceService` reads evidence by kind, per commit, and full history — all backend-less and cached. Regenerated TS bindings.

**Tooling & CI**
- `tools/evidence/publish_evidence.py`: computes a digest (informational), uploads the artifact to IPFS (pluggable via `TANSU_IPFS_UPLOAD_COMMAND`, or Filebase), and calls `set_evidence`. Supports pinning a directory as a bundle.
- `.github/workflows/evidence.yml`: generates an SBOM and vulnerability scan with Trivy, produces a Sigstore build-provenance attestation, **pins the attestation bundle + trusted root for offline verification**, uploads each to IPFS, and records the CIDs on-chain.
- Makefile targets + docs (`website/docs/developers/evidence.mdx`).

### Design notes
- **No on-chain `digest` field.** An IPFS CIDv1 is itself a content-addressed multihash, so fetching by CID is self-verifying — a separate digest would be redundant. The tooling still prints a `sha256` for human-readable logs.
- **Attestations are decentralized.** We pin the actual Sigstore bundl, so it verifies offline from IPFS alone via `gh attestation verify --bundle … --custom-trusted-root …`.
- **Storage-key safety.** New `ProjectKey` variants are name-tagged in Soroban storage, so existing deployed project/DAO/vote data is unaffected.

### Verifying evidence
```bash
# fetch the CID from chain
stellar contract invoke --network testnet --id "$TANSU_CONTRACT_ID" \
  -- get_evidence --project_key "$PK" --commit_hash "$COMMIT" --kind Sbom
# download by CID (IPFS verifies content against the CID)
ipfs cat "$CID" > evidence.json
```


tupui:
This is great thanks a lot. I will take some time to review in the next days. Any questions on your end otherwise?

tupui:
Something I wanted to point out which I like and in the spirit of this https://docs.rs/radicle-artifact/latest/radicle_artifact/

There might be some good things to borrow. If you did not see, there is now a `.radicle` folder with a release script which is using this.

dapp/src/service/EvidenceService.ts
tupui, line 1: @Shadow-MMN when you get some time if you could review only the TS files that would be helpful. Thanks!

Septimus4:
> Something I wanted to point out which I like and in the spirit of this https://docs.rs/radicle-artifact/latest/radicle_artifact/
> 
> There might be some good things to borrow. If you did not see, there is now a `.radicle` folder with a release script which is using this.

I will read it and see in what way we are similar and/or different

Septimus4:
For me the main differences come down to three things:
- Sources/locations: they let an artefact have multiple retrieval locations, whereas we just rely on IPFS.
- Schema: we use typed kinds (Sbom/Cve/Attestation), while they go generic arbitrary artifacts with free-form metadata, so basically anything.
- Mutability: Radicle supports redaction; ours is append-only.
Overall though, the designs are very similar: both are content-addressed (CID) and both tie artifacts to a commit or tag, we always use a commit hash for simplicity, but it's the same idea.
One thing I'd actually argue is a plus on our side: because we rely on content-addressed IPFS and don't add a separate source/provenance layer, there's no extra trust assumption. There's nothing in the middle that could be altered or tampered with the CID always self-verifies, so anyone can check an artefact at any time.

But what we could do is:
- Free-form kind, allow anything. No tradeoff here, but is it useful ?
- A redaction/supersede marker, still append-only on-chain, but a way to flag an entry as retracted (with a reason) so we can correct bad evidence instead of just leaving it. Or just allow mutation but would weaken a little bit the security transparency setup.
- Naming artefacts within a kind, if a commit ever has more than one of the same kind (e.g. multiple SBOMs per package). No tradeoff here so I think I can do it

.github/workflows/evidence.yml
tupui, line 1: Nice but we instead should include these into existing workflows.

contracts/tansu/src/tests/test_evidence.rs
tupui, line 1: Please move tests close to the commit ones. We could rename the test file `test_commit` to `test_versioning` maybe.

contracts/tansu/src/events.rs
```diff
@@ -27,6 +29,20 @@ pub struct Commit {
...
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct EvidenceSet {
+    #[topic]
+    pub project_key: Bytes,
+    pub commit_hash: String,
+    pub kind: EvidenceKind,
+    pub cid: String,
+    /// Zero-based index of this entry in the append-only history for
+    /// (project_key, commit_hash, kind). Lets consumers reconstruct the full
+    /// timeline from events alone.
```
tupui: This is more for the docs

contracts/tansu/src/contract_versioning.rs
```diff
@@ -5,6 +5,67 @@ use crate::{Tansu, TansuArgs, TansuClient, TansuTrait, VersioningTrait, errors,
...
 const REGISTER_COLLATERAL: i128 = 5 * 10_000_000;
 
+// Durability for evidence entries. Evidence is meant to be a long-lived,
+// backend-less historical record, so each entry is bumped towards the network
+// maximum persistent TTL on write (and can be re-bumped permissionlessly via
+// `bump_evidence`). Values stay below the protocol `max_entry_ttl` (6_312_000
+// ledgers, ~1 year) to avoid overflowing it. If rent still lapses, persistent
+// entries are archived (not deleted) and remain restorable on-chain.
+const EVIDENCE_BUMP_THRESHOLD: u32 = 2_592_000; // re-extend when remaining TTL drops below ~150 days
+const EVIDENCE_BUMP_AMOUNT: u32 = 6_000_000; // extend live window to ~347 days
```
tupui: We don't care about doing this. We store with persistant and if people want to restore, they do and pay the rent. I have some plans to handle storage better in general and will do later. For now just do like the rest of the storage basically.

contracts/tansu/src/contract_versioning.rs
```diff
@@ -220,6 +281,169 @@ impl VersioningTrait for Tansu {
...
+
+    /// Extend the TTL of a historical evidence entry, keeping it alive on-chain.
+    ///
+    /// Permissionless on purpose: anyone may pay to preserve a project's evidence
+    /// history. It can only extend rent, never modify or read out the data.
+    ///
+    /// # Panics
+    /// * If the project doesn't exist
+    /// * If no evidence exists at that index
+    fn bump_evidence(
```
tupui: Not needed as said above.

contracts/tansu/src/contract_versioning.rs
```diff
@@ -5,6 +5,67 @@ use crate::{Tansu, TansuArgs, TansuClient, TansuTrait, VersioningTrait, errors,
...
+// maximum persistent TTL on write (and can be re-bumped permissionlessly via
+// `bump_evidence`). Values stay below the protocol `max_entry_ttl` (6_312_000
+// ledgers, ~1 year) to avoid overflowing it. If rent still lapses, persistent
+// entries are archived (not deleted) and remain restorable on-chain.
+const EVIDENCE_BUMP_THRESHOLD: u32 = 2_592_000; // re-extend when remaining TTL drops below ~150 days
+const EVIDENCE_BUMP_AMOUNT: u32 = 6_000_000; // extend live window to ~347 days
+
+impl Tansu {
+    /// Panic with `InvalidKey` if the project does not exist.
+    fn require_project(env: &Env, project_key: &Bytes) {
```
tupui: I would not think this is really needed. My reading is that you added this to have a nicer error message. I think it's ok not to do because the code would anyway panic as not finding an evidence. The reason for not finding it is check your inputs. As long as we do protect well the writting to only allow existing projects to write an evidence then it's ok to me.

contracts/tansu/src/contract_versioning.rs
```diff
@@ -220,6 +281,169 @@ impl VersioningTrait for Tansu {
...
+        Self::require_project(&env, &project_key);
+        Self::evidence_count(&env, &project_key, &commit_hash, &kind)
+    }
+
+    /// Get a specific historical evidence entry by its zero-based index.
+    ///
+    /// # Panics
+    /// * If the project doesn't exist
+    /// * If no evidence exists at that index
+    fn get_evidence_at(
```
tupui: I am not sure we need all that in terms of API.

What I propose instead is to store a vec of evidence per kind. Bound the vec to max 10 entries for instance. Not sure exactly what would be a good number. I am not even sure we need to separate by kind. In the end I am expecting things to be relatively light in terms of usage. Having a few evidences per commits at most. Meaning we could have a single list.
`(project_key, commit_hash) -> [{CID, kind}, {CID, kind}, ...]`

What do you think?

Also on the API, what can be done via an indexer should not really be in the contract IMO. Unless we really need it.

tupui (commented):
@Septimus4 I had some time to have a look and left some thoughts. Thanks again!

tools/evidence/publish_evidence.py
tupui, line 1: I mean you did not have to write this. Sort of up to people to manage this. Like Radicle does not have a way with radicle-attestation to upload thing either.

I think a simple bash script would have suffice as a helper otherwise. Because in the end it's calling 2 CLI tools, curl and then making a smart contract call with the data. No need to overcomplexify things.

dapp/src/service/EvidenceService.ts
Shadow-MMN: Just seeing this today, my bad 😥
Will add my review, when he is done with the review you gave him

contracts/tansu/src/contract_versioning.rs
Septimus4: I don't think there is any better and worst choice here, we just need to decide, it will heavily depend on user adoption adn usage, I usually prefer to prepare for heavy usage because it's usually the more complex one and it's always easier to change to lighter. We can start with 10 and always make it bigger later.

Same for the feature scope, we can definitely remove all get_* logic and keep it lighter, the indexer does have all this data

tools/evidence/publish_evidence.py
Septimus4: Frankly I did it for me and dev purposes to make it simpler knowing it could also be use downstream by maintainers. But I can make it simpler

contracts/tansu/src/contract_versioning.rs
tupui: ok then let's go simple 👍 Thanks!

contracts/tansu/src/contract_versioning.rs
```diff
@@ -220,6 +225,96 @@ impl VersioningTrait for Tansu {
...
+        kind: types::EvidenceKind,
+        cid: String,
+    ) {
+        Tansu::require_not_paused(env.clone());
+
+        crate::auth_maintainers(&env, &maintainer, &project_key);
+
+        if commit_hash.is_empty() || cid.is_empty() {
+            panic_with_error!(&env, &errors::ContractErrors::InvalidEvidence);
+        }
```
tupui, line 265: Not very helpful like this but better than nothing so ok. We could check that these are of the correct size and using ascii chars only. If we did we could do that elsewhere too. But I will make a separate issue for this.

contracts/tansu/src/contract_versioning.rs
```diff
@@ -220,6 +225,96 @@ impl VersioningTrait for Tansu {
...
+
+        let mut history: Vec<types::Evidence> = storage.get(&key).unwrap_or_else(|| Vec::new(&env));
+        history.push_back(types::Evidence {
+            cid: cid.clone(),
+            created_at: env.ledger().timestamp(),
+        });
+        // Keep only the most recent MAX_EVIDENCE entries; the full timeline lives in events.
+        while history.len() > MAX_EVIDENCE {
+            history.remove(0);
+        }
```
tupui, line 279: Not a if? Why a while?

Alternatively we could have used a slice if we had reversed the order. Both approaches are fine with me.

contracts/tansu/src/tests/test_versioning.rs
tupui, line 1: A bit verbose and some tests could be bundled together but it's fine. Optinal comment.

.github/workflows/sbom.yml
```diff
@@ -95,3 +111,100 @@ jobs:
...
+
+      - name: Install Stellar CLI
+        run: |
+          curl -fsSL -o stellar-cli.tar.gz \
+            "https://github.com/stellar/stellar-cli/releases/download/v${STELLAR_CLI_VERSION}/stellar-cli-${STELLAR_CLI_VERSION}-x86_64-unknown-linux-gnu.tar.gz"
+          tar -xzf stellar-cli.tar.gz
+          sudo install -m 0755 stellar /usr/local/bin/stellar
+          stellar --version
+
+      - name: Generate evidence artifacts
```
tupui: No no, we don't need to redo things in this action. We already have all the artifacsts we want and just need to call the contract with these. Same think for the release.yml workflow. I expect just a quick call to the CLI, not much more.

tupui (commented):
Thanks, I left some minor things. This is looking good now on the contract side to me 🚀

contracts/tansu/src/contract_versioning.rs
Septimus4: Yeah if works fine today since each call only adds one, so it can't overshoot by more than one. I kept the while on purpose though.
If we ever lower MAX_EVIDENCE in an upgrade, a while trims existing oversized vecs straight back down to the cap, whereas an if would just stay stuck (removes one, adds one, never converges). It's bounded so there's no real cost to looping.
It woudl also avoid being stuck if for any reason we would get in a bad state with more than 10. So basically it's just more robust whith no real tradeoffs

tupui: Actually a good point 👍 Great thinking

tupui (approved):
Alright, getting this in. Thanks again @Septimus4 for working on this. Really awesome addition. Now I need to push for people to leverage this in projects 🚀

tupui:
I made this follow up #204

PR #187, opened 2026-06-23
feat: Add Git identity binding during member registration

### LINKED ISSUE #16: Add Git identity binding during member registration
xref https://github.com/tupui/soroban-versioning/issues/217

Shadow-MMN:
### Summary
Implements SEP-53 Git identity binding during member registration. Members can now optionally link a Git (GitHub/GitLab) identity to their on-chain account by providing a message signed with their Ed25519 SSH key.

### Smart Contract Changes ( `contracts/tansu/` )
-  add_member : Added optional parameters  git_identity  (String),  git_pubkey  (BytesN<32>),  git_msg  (Bytes),  git_sig  (BytesN<64>). When provided, verifies the Ed25519 signature against the pre-computed SSH signed payload before persisting.
-  update_member : Added optional  git_identity  and  git_pubkey  parameters.
-  Member  struct ( types.rs ): Added  git_identity: Option<String>  and  git_pubkey: Option<BytesN<32>>  fields.
-  GitIdentityAdded  event: Emitted when a Git identity is bound.
- New error codes:  InvalidGitSignature  (700),  InvalidGitIdentity  (701).
-  verify_git_signature : Internal helper using  env.crypto().ed25519_verify() .

### Frontend Changes ( `dapp/` )
-  GitVerification.tsx : New component for Git identity verification in the registration flow.
-  sshSignature.ts : Utility for constructing and verifying SEP-53 SSH signed messages.
- Playwright tests ( git-identity-flows.spec.ts ): End-to-end flows for git identity binding.
- Unit tests ( sshSignature.test.ts ): Coverage for the SSH signature utility.
- Updated contract bindings ( packages/tansu/ ): Regenerated to reflect new  add_member / update_member  signatures and error codes.

####Testing
- Rust: 62 tests pass, including 9 new tests in  test_git_identity.rs  covering:
- Happy path with GitHub/GitLab handles
- Missing pubkey, missing message, bad signature, wrong pubkey
- Duplicate member registration with git identity
- Backward compatibility (registration without git identity)
- Frontend: 235 unit tests pass, plus new Playwright E2E flows

### Visual Implementation 
Screenshot 2026-06-23 at 19 10 09
Screenshot 2026-06-23 at 19 10 17
Screenshot 2026-06-23 at 19 10 35
Screenshot 2026-06-23 at 19 15 19
Screenshot 2026-06-23 at 21 13 30

Closes #16


contracts/tansu/Cargo.toml
```diff
@@ -13,3 +13,6 @@ soroban-sdk = { workspace = true }
 
 [dev-dependencies]
 soroban-sdk = { workspace = true, features = ["testutils", "experimental_spec_shaking_v2"] }
+ed25519-dalek = "2.2"
```
tupui, line 16: Can't we just use the Soroban SDK?

contracts/tansu/src/errors.rs
```diff
@@ -48,4 +48,8 @@ pub enum ContractErrors {
     UpgradeError = 601,
     ContractValidation = 602,
     CollateralError = 603,
+
+    // Git identity (700-799)
```
tupui: Not a new section

contracts/tansu/src/events.rs
```diff
@@ -150,3 +150,10 @@ pub struct ConflictOfInterestUpdated {
     pub maintainer: Address,
     pub changed: Vec<Address>,
 }
+
+#[contractevent]
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct GitIdentityAdded {
```
tupui: Not needed, update the existing event

contracts/tansu/src/lib.rs
```diff
@@ -54,9 +54,24 @@ pub trait TansuTrait {
...
 pub trait MembershipTrait {
-    fn add_member(env: Env, member_address: Address, meta: String);
+    #[allow(clippy::too_many_arguments)]
+    fn add_member(
+        env: Env,
+        member_address: Address,
+        meta: String,
+        git_identity: Option<String>,
+        git_pubkey: Option<BytesN<32>>,
+        git_msg: Option<Bytes>,
```
tupui: The update method and this one must have the same API as they do the same.

contracts/tansu/src/contract_membership.rs
```diff
@@ -62,6 +123,13 @@ impl MembershipTrait for Tansu {
             None => panic_with_error!(&env, &errors::ContractErrors::UnknownMember),
             Some(mut member) => {
                 member.meta = meta;
+                // Only update git fields if explicitly provided
+                if let Some(ref identity) = git_identity {
+                    member.git_identity = Some(identity.clone());
```
tupui: We must do the same as with the membership creating and check the signature. Otherwise someone can just pretend here to have any identity by calling this function after the creation

contracts/tansu/src/contract_membership.rs
```diff
@@ -283,6 +351,17 @@ impl MembershipTrait for Tansu {
...
 }
 
+/// Verify a Git identity binding.
+///
+/// The `msg` parameter is the pre-computed SSH signed payload from the dapp:
+///   "SSHSIG" + wire("file") + wire("") + wire("sha512") + wire(SHA-512(envelope))
+///
+/// The dapp computes SHA-512 and constructs the SSH-wire format off-chain.
+/// The contract simply verifies the Ed25519 signature against the signed payload.
+fn verify_git_signature(env: &Env, pubkey: &BytesN<32>, msg: &Bytes, sig: &BytesN<64>) {
```
tupui: Useless function as it is. But since you will also reconstruct the message here, it will be useful. That means the functions must be able to get the needed parameters.

Also make sure the comments are handwritten. Right now it's bad.

contracts/tansu/src/types.rs
```diff
@@ -56,6 +56,10 @@ pub struct ProjectBadges {
 pub struct Member {
     pub projects: Vec<ProjectBadges>,
     pub meta: String,
+    /// Optional Git identity bound to this member (e.g. "github:alice").
```
tupui: Redundant to say optional.

contracts/tansu/src/types.rs
```diff
@@ -56,6 +56,10 @@ pub struct ProjectBadges {
...
+    pub git_identity: Option<String>,
+    /// Ed25519 public key (32 bytes) registered on the Git account.
```
tupui: Redundant to say 32 bytes

contracts/tansu/src/tests/test_git_identity.rs
```diff
@@ -0,0 +1,335 @@
...
+    let signing_key = SigningKey::from_bytes(&secret);
+    let verifying_key = signing_key.verifying_key();
+    let pubkey_bytes = verifying_key.to_bytes();
+
+    let nonce = "feedbeefcafebabedeadbeef01234567";
+    let signing_account = "GAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAWHF";
+
+    // Build the 5-line SEP-53 envelope
+    let envelope = [
+        "Stellar Signed Message\nTest SDF Network ; September 2015\n",
```
tupui: We don't want to have the network passphrase in the message. This is useless.

I am also not sure a nonce is needed since we would use the member address and check that this address signed the transaction. So someone else cannot reuse the signature. And we don't have any double spend problematic to handle.

Also do not pass the SEP-53 prefix when making onchain call, this must be declared as a cst

contracts/tansu/src/contract_membership.rs
```diff
@@ -1,21 +1,47 @@
...
+    ///   1. ed25519_verify(git_pubkey, git_msg, git_sig) where git_msg is the
+    ///      pre-computed SSH signed payload from the dapp
+    ///
+    /// Only `git_identity` and `git_pubkey` are persisted on the `Member`.
+    ///
     /// # Arguments
     /// * `env` - The environment object
     /// * `member_address` - The address of the member to add
-    /// * `meta` - Metadata string associated with the member (e.g., IPFS hash)
+    /// * `meta` - Metadata string (e.g., IPFS hash)
```
tupui: Do not change my docs. I wrote all that by hand and it's better than this.

contracts/tansu/src/tests/test_git_identity.rs
```diff
@@ -0,0 +1,335 @@
...
+        signing_account,
+        "\n",
+        nonce,
+        "\ntansu-bind|deadbeef|",
```
tupui: What is this?

Really rethink the message structure. It's ok to have delimiters and a prefix (but shorter than this), but then really put the prefix at the beginning and be consistent with the delimiters usage.

contracts/tansu/src/tests/test_git_identity.rs
tupui, line 1: We don't want another file, just add these to the existing test file about membership.

tupui (changes requested):
Good start, I left some comments.

I will check the UI afterwards. There will be a need to have a generic option and not just GitLab or GitHub. And if there is a binding of identity we need to see that information on a member info

contracts/tansu/Cargo.toml
Shadow-MMN: I looked into that. My understanding is that the Soroban SDK exposes `env.crypto().ed25519_verify()` for verification and hashing functions like `sha256()`, but I couldn't find host APIs for generating Ed25519 keypairs, producing signatures, or computing SHA-512, which the tests use to simulate the `ssh-keygen -Y sign` flow. That's why I added `ed25519-dalek` and `sha2` as `dev-dependencies` only.

But if there is a better approach, you can tell me let me follow that route.

Shadow-MMN:
@tupui for taking your time to review my changes, will be pushing a new one soon to address everything you said.

contracts/tansu/Cargo.toml
tupui: Ok it's fine 👍

dapp/packages/tansu/package.json
```diff
@@ -8,7 +8,7 @@
     "build": "tsc"
   },
   "dependencies": {
-    "@stellar/stellar-sdk": "^16.0.0",
+    "@stellar/stellar-sdk": "^14.5.0",
```
tupui: No this is a mistake

contracts/tansu/src/contract_membership.rs
```diff
@@ -283,6 +380,35 @@ impl MembershipTrait for Tansu {
     }
 }
 
+/// Wire-encode a bytes slice for the SSH signature format:
+/// uint32(len) + bytes.
+fn wire_bytes(env: &Env, data: &[u8]) -> Bytes {
```
tupui: You are overthinking this, have a look at this function https://github.com/Consulting-Manao/chimpdao-contracts/blob/5dd95a6728018cf0b1b82dd4be8f2693adb0d9bf/contracts/nfc-nft/src/contract.rs#L325

contracts/tansu/src/contract_membership.rs
```diff
@@ -1,21 +1,50 @@
...
-    Address, Bytes, Env, I256, InvokeError, String, Symbol, Vec, contractimpl, panic_with_error,
-    vec,
+    Address, Bytes, BytesN, Env, I256, InvokeError, String, Symbol, Vec, contractimpl,
+    panic_with_error, vec,
 };
 
+/// NS used in the SSHSIG payload when verifying git signatures.
+const SSHSIG_NAMESPACE: &[u8] = b"file";
+/// Hash algorithm used in the SSHSIG payload.
+const SSHSIG_HASH_ALG: &[u8] = b"sha512";
```
tupui: No, we don't want these.

contracts/tansu/src/contract_membership.rs
```diff
@@ -283,6 +380,35 @@ impl MembershipTrait for Tansu {
...
+
+/// Verify a Git identity binding by reconstructing the SSH signed payload
+/// from the provided SHA-512 hash and checking the Ed25519 signature.
+///
+/// The reconstructed payload matches what `ssh-keygen -Y sign` produces:
+///   "SSHSIG" + wire(namespace) + wire("") + wire(hash_alg) + wire(hash)
+///
+/// The hash is SHA-512 of the message:
+///   `TANSU_BIND\n<member_address>\n<git_identity>`
+fn verify_git_signature(env: &Env, pubkey: &BytesN<32>, git_hash: &BytesN<64>, sig: &BytesN<64>) {
```
tupui: No this is not how this should work. You also removed the SEP-53 part now. Please don't use AI for the smart contract part. It's clearly doing no good. It's like none of my comments are really addressed and now it's bad differently.

We want (in pseudo code):

`msg = "Stellar Signed Message:\n" + member_address + git_pubkey + git_identity`

So the owner of both `member_address` and `git_pubkey` needs to sign this message which is `git_sig` (`git_sig = sign(msg)`). In the contract we must check that `member_address` is authorized and that `git_sig` has been signed by the git private key. The owner of the git private key would only sign a message that has it's own `member_address` in it. This is how it's safe in this context. So someone cannot reuse the signature later on as they would not be able to sign with `member_address`.

tupui (changes requested):

Shadow-MMN:
The one remaining limitation is that `ed25519_verify` on failure still bubbles as a host error rather than a clean ContractError, Soroban SDK 26 doesn't expose `env.try()` to catch host function panics. The `try_add_member` wrapper does catch it as `Err(HostError)`, so the frontend can distinguish success from failure.

That said, I am still working on the Generic git provider option

Shadow-MMN:
#### visual for the radicle implementation
Screenshot 2026-07-01 at 12 24 15

Shadow-MMN:
For the noble/curves package, It's used at `GitVerification.tsx:209` to verify `Ed25519` signatures client-side.

contracts/tansu/src/contract_membership.rs
```diff
@@ -29,26 +48,70 @@ impl MembershipTrait for Tansu {
         {
             panic_with_error!(&env, &errors::ContractErrors::MemberAlreadyExist)
         } else {
+            // Validate and verify Git identity if provided
+            if let Some(ref identity) = git_identity {
+                let pubkey = git_pubkey.as_ref().unwrap_or_else(|| {
```
tupui: Why the ref and as_ref?

Also we don't need to put all that under an else. It was ok before as we had barely any code but now it just some nesting for nothing.

contracts/tansu/src/contract_membership.rs
```diff
@@ -1,21 +1,40 @@
...
 #[contractimpl]
 impl MembershipTrait for Tansu {
     /// Add a new member to the system with metadata.
     ///
+    /// Optionally binds a Git identity using message signing.
+    /// When `git_identity` is `Some`, the message is reconstructed and
+    /// verified using ed25519_verify:
+    ///   msg = "Stellar Signed Message:\n" || member_address || git_pubkey || git_identity
+    ///
+    /// Only `git_identity` and `git_pubkey` are persisted.
```
tupui: This is an implementation detail and should not be public API facing, this is also duplicate from the verify function itself.

contracts/tansu/src/contract_membership.rs
```diff
@@ -283,6 +368,65 @@ impl MembershipTrait for Tansu {
...
+    sig: &BytesN<64>,
+) {
+    // Build the raw message
+    let mut msg = Bytes::new(env);
+    msg.append(&Bytes::from_slice(env, b"Stellar Signed Message:\n"));
+    msg.append(&member_address.to_string().into());
+    msg.append(&Bytes::from_slice(env, &git_pubkey.to_array()));
+    msg.append(&git_identity.clone().into());
+
+    // SHA-256 of the message
```
tupui: Comments like these are useless. It's in the name of the function you call.

contracts/tansu/src/contract_membership.rs
```diff
@@ -283,6 +368,65 @@ impl MembershipTrait for Tansu {
...
+    let hash_bytes: [u8; 32] = hash.to_array();
+
+    // Build the SSHSIG tosign payload — the exact payload that
+    // `ssh-keygen -Y sign -O hashalg=sha256 -n file` signs:
+    //   "SSHSIG" + string("file") + string("") + string("sha256") + string(SHA-256(msg))
+    let mut tosign = Bytes::new(env);
+    tosign.append(&Bytes::from_slice(env, b"SSHSIG"));
+    // string("file")
+    tosign.append(&Bytes::from_slice(env, &4u32.to_be_bytes()));
+    tosign.append(&Bytes::from_slice(env, b"file"));
```
tupui: You know that you can use the flag `-n` to actually have a namespace instead of `file` or `email`. Please carefully read the docs.

contracts/tansu/src/contract_membership.rs
```diff
@@ -283,6 +368,65 @@ impl MembershipTrait for Tansu {
...
 }
 
+/// Verify a Git identity binding by checking the Ed25519 signature over
+/// a plain message containing the member address, public key, and identity.
+///
+/// The signed message is:
+///   msg = "Stellar Signed Message:\n" || member_address || git_pubkey || git_identity
+///
+/// (raw byte concatenation via Bytes::append — no length prefixes, no
+/// namespace tags, no intermediate hashing step).
```
tupui: Useless comment

contracts/tansu/src/contract_membership.rs
```diff
@@ -283,6 +368,65 @@ impl MembershipTrait for Tansu {
...
+///
+/// The member_address is embedded in the message so that a signature created
+/// for one Stellar account cannot be replayed to bind the same git identity
+/// to a different account.
```
tupui: Remove as technically not correct. Everyone can still copy that signature. The replay is prevented because the G account must be the one that actually also sign the contract call. That is the protection. You know that the owner of G is the owner of the SSH/GPG key

dapp/packages/scf-membership/package.json
```diff
@@ -8,7 +8,7 @@
     "build": "tsc"
   },
   "dependencies": {
-    "@stellar/stellar-sdk": "^16.0.0",
+    "@stellar/stellar-sdk": "^14.5.0",
```
tupui: Still no

dapp/src/utils/sshSignature.ts
```diff
@@ -0,0 +1,169 @@
...
+ *   uint32              — version (1)
+ *   string              — public key (key-type + key-material)
+ *   string              — namespace
+ *   string              — reserved
+ *   string              — hash algorithm
+ *   string              — signature (key-type + signature-material)
+ *
+ * Returns the 64-byte Ed25519 signature or null.
+ */
+function decodeSshSignaturePayload(raw: Uint8Array): Uint8Array | null {
```
tupui: Really needed to have all these things? I thought now we could verify the signature without tricks.

tupui (changes requested):

Shadow-MMN:
@tupui , thank you for taking your time to review my code
For everything you pointed out one by one , here are the steps I took :
- Replaced `ref identity` with `&git_identity`, flattened the else by using an early return on the duplicate check. No more unnecessary nesting.
- Stripped the message format implementation detail from the public-facing doc on `add_member`. It now says what it does, not how.
- Removed the `// SHA-256 of` the message comment, the function is already clear.
- Made the SSHSIG namespace configurable. Added `git_namespace: Option<String>` to `add_member/update_member` which defaults to "file" when `None`, so existing callers don't need to change.
- Removed the raw byte concatenation via Bytes::append line from the verify_git_signature doc.
-  Fixed the replay-protection comment. The address is embedded to tie the SSH key proof to the Stellar account, not as replay protection — that comes from require_auth() on the caller.
- Reverted `@stellar/stellar-sdk` back to ^16.0.0 in `scf-membership/package.json` and ran bun install to update the lockfile .
- Stripped out the entire `SSH` wire-format parsing (`decodeSshSignaturePayload`, `extractFromArmored`). The frontend now only accepts raw `base64-encoded` `64-byte signatures`, and the `ssh-keygen` instructions show how to pipe the output to extract the raw signature with `tail | head | base64 -d | tail -c 64 | base64`.

Additional things I did which I saw while working are :
- Fixed Python signing instructions in GitVerification.tsx to build SSHSIG payload before signing instead of signing the raw message directly.
- Fixed Node.js signing instructions in GitVerification.tsx identically.

contracts/tansu/src/contract_membership.rs
```diff
@@ -1,21 +1,38 @@
...
+    /// * If git params are incomplete (identity, key, sig must be all Some or None)
+    /// * If the signature verification fails
+    fn add_member(
+        env: Env,
+        member_address: Address,
+        meta: String,
+        git_identity: Option<String>,
+        git_pubkey: Option<BytesN<32>>,
+        git_sig: Option<BytesN<64>>,
+        git_namespace: Option<String>,
```
tupui: No do not add namespace, this is not what I asked. You should fix this to `tansu` or else and in the CLI add `-n tansu` that's it.

contracts/tansu/src/contract_membership.rs
```diff
@@ -28,27 +45,60 @@ impl MembershipTrait for Tansu {
...
-        } else {
-            let member = types::Member {
-                projects: Vec::new(&env),
-                meta,
-            };
-            env.storage().persistent().set(&member_key_, &member);
+        }
+
+        if let Some(identity) = &git_identity {
+            let pubkey = git_pubkey.as_ref().unwrap_or_else(|| {
```
tupui: Still no, all of this has to be simplified with the use of `is_some()` like we have in other places.

contracts/tansu/src/contract_membership.rs
```diff
@@ -62,9 +112,37 @@ impl MembershipTrait for Tansu {
             None => panic_with_error!(&env, &errors::ContractErrors::UnknownMember),
             Some(mut member) => {
                 member.meta = meta;
+
+                if let Some(identity) = &git_identity {
+                    let pubkey = git_pubkey.as_ref().unwrap_or_else(|| {
```
tupui: Same here

contracts/tansu/src/contract_membership.rs
```diff
@@ -283,6 +361,58 @@ impl MembershipTrait for Tansu {
...
+    msg.append(&Bytes::from_slice(env, &git_pubkey.to_array()));
+    msg.append(&git_identity.clone().into());
+
+    let hash = env.crypto().sha256(&msg);
+    let hash_bytes: [u8; 32] = hash.to_array();
+
+    let ns_bytes: soroban_sdk::Bytes = match git_namespace {
+        Some(ns) => ns.clone().into(),
+        None => soroban_sdk::Bytes::from_slice(env, b"file"),
+    };
```
tupui: As said above, use `tansu` as a fixed value.

contracts/tansu/src/contract_membership.rs
```diff
@@ -283,6 +361,58 @@ impl MembershipTrait for Tansu {
...
+
+    let mut tosign = Bytes::new(env);
+    tosign.append(&Bytes::from_slice(env, b"SSHSIG"));
+    tosign.append(&Bytes::from_slice(env, &ns_bytes.len().to_be_bytes()));
+    tosign.append(&ns_bytes);
+    tosign.append(&Bytes::from_slice(env, &0u32.to_be_bytes()));
+    tosign.append(&Bytes::from_slice(env, &6u32.to_be_bytes()));
+    tosign.append(&Bytes::from_slice(env, b"sha256"));
+    tosign.append(&Bytes::from_slice(env, &32u32.to_be_bytes()));
```
tupui: All the above is fixed and we can do this in one line then no?

Shadow-MMN:
I don't understand why the website-dapp linting is failing, it is passing locally 

Screenshot 2026-07-03 at 06 39 14

dapp/src/service/FlowService.ts
```diff
@@ -31,6 +31,12 @@ interface JoinCommunityFlowParams {
   memberAddress: string;
   profileFiles: File[];
   onProgress?: (step: number) => void;
+  gitIdentity?: {
+    gitIdentity: string;
+    gitPubkey: Buffer;
+    gitSig: Buffer;
+    gitNamespace?: string;
```
tupui: Leftover

dapp/src/service/FlowService.ts
```diff
@@ -124,11 +130,17 @@ async function createSignedProposalTransaction(
...
+ * Create and sign an add member transaction, optionally with Git identity binding.
  */
 async function createSignedAddMemberTransaction(
   memberAddress: string,
   meta: string,
+  gitIdentity?: {
+    gitIdentity: string;
+    gitPubkey: Buffer;
+    gitSig: Buffer;
+    gitNamespace?: string;
```
tupui: Here too

dapp/src/service/FlowService.ts
Shadow-MMN: removed them now

dapp/src/service/FlowService.ts
Shadow-MMN: done also.

contracts/tansu/src/contract_membership.rs
Shadow-MMN: Removed namespace .

contracts/tansu/src/contract_membership.rs
Shadow-MMN: Removed also.

contracts/tansu/src/contract_membership.rs
Shadow-MMN: We can't reduce it to a single line because the hash is computed at runtime, but I am thinking of extracted the static 33-byte SSHSIG prefix into a `const`, so the `tosign` construction can now be just two operations: 
-`Bytes::from_slice` for the static prefix 
-`append` for the runtime SHA-256 hash.

contracts/tansu/src/contract_membership.rs
Shadow-MMN: Done also.

contracts/tansu/src/contract_membership.rs
Shadow-MMN: Used `is_some()` just as have used in other places.

tupui (approved):
Thanks let's get this one in 🙌

Shadow-MMN:
> Thanks let's get this one in 🙌

Thank you very much for all your reviews.
Will start working on this today 
https://github.com/Consulting-Manao/tansu/pull/189#issuecomment-4852129319

PR #188, opened 2026-06-25
tansu commit 

### LINKED ISSUE #179: [bug] Join Community: editable Member Address lets you submit an unauthorizable join (only self-registration works)
### Summary
The "Join the Community" modal exposes the **Member Address** as an editable, validated input, and when a wallet is connected it submits the join for whatever is typed. But the contract's `add_member` requires the member to authorize their own registration (`member_address.require_auth()`), so any address other than the connected wallet produces a transaction that can never be authorized and fails with a cryptic error.

### Steps to Reproduce
1. Connect wallet A.
2. Click **Join** to open the Join the Community modal.
3. In the **Member Address** field, replace the prefilled value with a different valid address `B` (e.g. paste a teammate's `G…` address).
4. Click **Join**.

### Expected Behavior
The join either succeeds for a member you control, or the UI clearly prevents entering an address you can't authorize.

### Actual Behavior
The transaction is built for address `B` but signed by wallet `A`, so `member_address.require_auth()` fails (and the source account mismatches). The join fails with a confusing error, even though the UI accepted the input as valid.

### Additional Inconsistency
The connected vs. disconnected paths use different addresses for the same form:
- **Wallet connected** → uses the typed `address` field.
- **Not connected** → after `authModal()`, uses the wallet's `connectedAddress` and ignores the typed field.

So the same form behaves differently depending on connection state.

### Impact
Users who edit or paste a different address (the field invites it with the placeholder "Write the address as G…") hit an unexplained failure. Confusing UX; the UI permits an action that can never succeed.

Please check this video.

https://github.com/user-attachments/assets/e4e0d083-376e-48da-bae1-5fb169da2f9c

Cerome360:
Closes #179

  Summary & What Changed
  The "Join the Community" modal previously allowed users to type an arbitrary member address into the input field. Because the underlying contract 
  add_member  function requires the member to authorize their own registration ( member_address.require_auth() ), typing an address different from the
  connected wallet resulted in a cryptic transaction failure.
  This PR:

  • Disables the "Member Address" input field so it's strictly read-only.
  • Updates the  handleJoin  connected flow to enforce building the transaction strictly with the  publicKey .
  • Updates the disconnected flow to explicitly tie the  address  state to the returned  connectedAddress  immediately upon wallet connection.
  • Updates the  happy-flows.spec.ts  E2E test to align with the disabled input field behavior.
  Key Design Decisions

  • The easiest, most robust way to guarantee the source account matches the member address is tying the form directly to the wallet connection payload.
  Validating an arbitrary typed address was dropped because manual typing is fundamentally incompatible with the self-authorization required by the
  contract.

  Acceptance Criteria Checklist

  [✓] The UI clearly prevents entering an address the user cannot authorize (input is disabled).
  [✓] The connected vs. disconnected paths use the exact same logic (bound to the connected wallet).
  [✓] Typechecks and lint pass locally.

  Test & Coverage Output

  • E2E playwright tests run and pass cleanly ( 33 passed (1.9m) ).
  • Local typecheck ( astro check ) passes cleanly ( 0 errors ).
  • Note: No component-level unit test coverage metrics exist in this repo for  JoinCommunityModal.tsx , but e2e suites explicitly verify the visual
  presence and modal flow.

  Security Note
  The address payload sent to the join flow is now securely derived straight from the wallet payload instead of trusting generic client-side text input.

  Follow-Ups

  • A broader review of other forms using the generic  <Input>  component might be warranted to ensure there are no other instances where user input
  incorrectly drives authorization payloads.

### DIFF
--- a/dapp/src/components/page/dashboard/JoinCommunityModal.tsx
+++ b/dapp/src/components/page/dashboard/JoinCommunityModal.tsx
@@ -3,7 +3,7 @@ import Input from "components/utils/Input";
 import Button from "components/utils/Button";
 import FlowProgressModal from "components/utils/FlowProgressModal";
 import { loadedPublicKey, setConnection } from "@service/walletService";
-import { validateStellarAddress, validateUrl } from "utils/validations";
+import { validateUrl } from "utils/validations";
 import SimpleMarkdownEditor from "components/utils/SimpleMarkdownEditor";
 
 interface ProfileImageFile {
@@ -104,9 +104,9 @@ const JoinCommunityModal: FC<{
     name.trim() || social.trim() || description.trim() || profileImage;
 
   const validateAddressField = (): boolean => {
-    const err = validateStellarAddress(address);
-    setAddressError(err);
-    return err === null;
+    // Address is read-only and provided by wallet; skip validation
+    setAddressError(null);
+    return true;
   };
 
   const validateSocialField = (): boolean => {
@@ -175,7 +175,7 @@ const JoinCommunityModal: FC<{
       try {
         setIsLoading(true);
         setStep(6);
-        await doJoinFlow(address || publicKey);
+        await doJoinFlow(publicKey);
       } catch (err: any) {
         console.error("Join community error:", err);
         setError(err?.message || "Something went wrong");
@@ -199,6 +199,9 @@ const JoinCommunityModal: FC<{
           detail: { address: connectedAddress },
         }),
       );
+
+      // Reflect the actual connected address in the UI
+      setAddress(connectedAddress);
 
       try {
         setIsLoading(true);
@@ -248,12 +251,9 @@ const JoinCommunityModal: FC<{
 
           <Input
             label="Member Address *"
-            placeholder="Write the address as G..."
+            placeholder="Address will be filled upon connection"
             value={address}
-            onChange={(e) => {
-              setAddress(e.target.value);
-              setAddressError(null);
-            }}
+            disabled={true}
             error={addressError}
           />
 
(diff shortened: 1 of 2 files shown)


dapp/src/components/page/dashboard/JoinCommunityModal.tsx
```diff
@@ -104,9 +104,9 @@ const JoinCommunityModal: FC<{
     name.trim() || social.trim() || description.trim() || profileImage;
 
   const validateAddressField = (): boolean => {
-    const err = validateStellarAddress(address);
-    setAddressError(err);
-    return err === null;
+    // Address is read-only and provided by wallet; skip validation
```
tupui, line 107: This is bad, this function is then useless. We want to have a function that still validates addresses if we change it's value. Or we should not even have an input field to begin with which is also a UX consideration.

dapp/src/components/page/dashboard/JoinCommunityModal.tsx
```diff
@@ -200,6 +200,9 @@ const JoinCommunityModal: FC<{
         }),
       );
 
+      // Reflect the actual connected address in the UI
```
tupui, line 203: Do not leave comments like these. LLM leftover usually.

dapp/src/components/page/dashboard/JoinCommunityModal.tsx
```diff
@@ -248,12 +251,9 @@ const JoinCommunityModal: FC<{
 
           <Input
             label="Member Address *"
-            placeholder="Write the address as G..."
+            placeholder="Address will be filled upon connection"
```
tupui, line 254: Why changing this? And again if we decide to not allow a change, then we don't even need to present a field to begin with.

tupui (commented):
Thanks for working on this. Please don't use a LLM to write down what you are doing.

tupui:
@Cerome360 You have a review

tupui:
Time is up @Cerome360 I am closing and un-assigning you.

PR #189, opened 2026-06-25
fix(governance): handle OutcomeContract XDR decode crash and add cache error cool down.

Shadow-MMN:
## Problem
 
The governance page at `/governance/?name=stellarregistry` shows a loading spinner indefinitely and never displays proposals. Two root causes:
 
1. **XDR decode crash in `@stellar/stellar-sdk` v16** : When a proposal has an `OutcomeContract` with `args` (like the `publish_hash` outcome), the SDK throws `TypeError: invalid type scSpecTypeVal specified for string value` (tracked as stellar/js-stellar-sdk#1178). Since `get_dao` returns all proposals at once, a single broken proposal takes down the entire page.
 
2. **Infinite retry loop** : When an RPC call fails, the cache layer saw `status === "error" && data === undefined` and immediately retried. Each retry failure triggered a snapshot change, another retry, etc, the loading spinner never resolved.
 
### 1. Fallback in `getProposals()` (ReadContractService.ts)
 
If `get_dao` fails (typically from the XDR decode crash), fetch proposals individually via `get_proposal` and skip any that still fail decodability. This allows the list view to load the remaining valid proposals rather than showing a blank page.
 
### 2. Error cooldown in the cache layer (cacheStore.ts / cacheHooks.ts)
 
When `fetchWithCache` encounters a network error, set `expiresAt` to `Date.now() + 30_000` (30s cooldown). The `shouldFetch()` check now returns `false` if the cooldown is still active, breaking the infinite retry loop.
 
### 3. E2E test fixes
 
- **Soroban RPC mock** (`mock.ts`) : v16 SDK sends all RPC methods `simulateTransaction`, `sendTransaction`, `getTransaction`) to the same `/soroban/rpc` endpoint with the method name in the JSON body. The mock now inspects the POST body instead of the URL path and returns proper JSON-RPC-formatted responses.
- **Badge test** (`regression-prevention.spec.ts`) : Bypassed Terms of Service modal (was blocking page interactions), used a valid Stellar address (`G.padEnd(56, "B")` fails base32 checksum validation in the SDK), and mocked `setBadges` via route interception to avoid invalid XDR encoding in the Soroban simulation response mock.

## Visuals
- ### Before 

https://github.com/user-attachments/assets/e2d5ab28-07e2-4c18-8a71-de4f534de959

- ### After 

https://github.com/user-attachments/assets/35ac3152-0818-466c-af87-a7dbe22c63aa

### DIFF
--- a/dapp/src/service/ReadContractService.ts
+++ b/dapp/src/service/ReadContractService.ts
@@ -12,32 +12,6 @@ import { queryKeys } from "./cache/cacheKeys";
 const TTL_4H = 4 * 60 * 60 * 1000;
 const TTL_1H = 60 * 60 * 1000;
 
-async function hydrateProposalFromDaoItem(
-  project_name: string,
-  project_key: Buffer,
-  daoProposal: Proposal,
-): Promise<Proposal> {
-  try {
-    return await fetchWithCache(
-      queryKeys.proposal.raw(project_name, Number(daoProposal.id)),
-      async () => {
-        const proposalRes = await Tansu.get_proposal({
-          project_key,
-          proposal_id: Number(daoProposal.id),
-        });
-        checkSimulationError(proposalRes);
-        return proposalRes.result as Proposal;
-      },
-      {
-        ttlMs: TTL_1H,
-      },
-    );
-  } catch {
-    // Keep list rendering resilient when a single hydration fails.
-    return daoProposal;
-  }
-}
-
 async function getProjectHash(): Promise<string | null> {
   const projectId = loadedProjectId();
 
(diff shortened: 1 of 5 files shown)


Shadow-MMN:
Please @chadoh, would you be willing to review my PR when you have the chance.

tupui (approved):
Getting this in, thanks!

tupui:
@Shadow-MMN I think we still have some issues with this one. See https://testnet.tansu.dev/governance/?name=registry

Shadow-MMN:
> @Shadow-MMN I think we still have some issues with this one. See https://testnet.tansu.dev/governance/?name=registry

Ok will look at it today

PR #218, opened 2026-07-28
feat/Integrate discussions on proposals

### LINKED ISSUE #215: Integrate discussions on proposals
See the details on Radicle

https://radicle.network/nodes/radicle.consulting-manao.com/rad:zssaAF91kxuquZmZCV2SiK2FNX6s/issues/850a9420a6a4ac1fc0f091677455764fce3ab5b0

Submit a PR for Drips Wave points referencing this. Can be a readme or docs fix if you wanted to actually submit the real work via a Radicle patch.

Oluwatos94:
## Integrate discussions on proposals

  Closes #215 · Radicle: `850a9420a6a4ac1fc0f091677455764fce3ab5b0`

  Adds a **Discussion** section to the proposal page that renders a discussion thread and an optional summary loaded
  from the proposal's IPFS directory.

  ### Changes
  - **`DiscussionSection.tsx`** (new), renders below the outcome: an optional summary block plus a collapsible thread
  (author, formatted timestamp, Markdown body, optional "View original" link), following the existing `ReadmeViewer`
  pattern (`markdown-to-jsx` + `DOMPurify`) with matching loading/empty/failed states.
  - **`ProposalService.ts`** new readers `fetchProposalDiscussionFromIPFS` (`discussion.json`) and
  `fetchProposalDiscussionSummaryFromIPFS` (`summary.md`), reusing the cached multi-gateway IPFS helpers.
  - **`resolveDiscussionCid()`** single seam for where discussion docs live; defaults to the proposal's own immutable
  IPFS directory (frozen during voting, no IPNS), so the attach mechanism (on-chain field vs. post-execution) can change
  in one place later.
  - **`ProposalPage.tsx`** / **`types/proposal.ts`** — wire the section in and add `DiscussionPost` + optional
  `discussionIpfs`.
  - **Docs** documented the proposal IPFS directory and the `discussion.json` / `summary.md` format in
  `governance.mdx`.

  ### Testing
  - 27/27 unit tests in `ProposalService.test.ts` ✓
  - Verified populated + empty states in the dev server

image


christian-rogobete:
Really glad to see this feature, thank you for building it! One thing I noticed while checking it against live data: the merge workflow in the proposals repo currently snapshots the PR discussion as a single markdown file (`proposal_discussion.md`, next to `proposal.md` and `outcomes.json` in the same directory CID). On a real Q2 proposal CID, `/discussion.json` and `/summary.md` return 404, while `/proposal_discussion.md` returns the content. So with the current uploader the section would show the empty state for all existing proposals.

Two possible ways to close the gap: update the uploader to also emit `discussion.json`, or add a markdown fallback that renders `proposal_discussion.md` when `discussion.json` is absent. The fallback has the advantage that every proposal produced by the current uploader becomes displayable without touching the proposals repo. The existing Q2 ones already contain `proposal_discussion.md`, and the Q3 ones will too when their PRs merge.

tupui:
Thanks for checking this out @christian-rogobete! I agree that we should just use the current file naming. For summary as this is a new thing it's ok to use any name and yes that has to be fully optional.

@Oluwatos94 Please check locally with the existing project `stellarpga`.

Oluwatos94:
> Really glad to see this feature, thank you for building it! One thing I noticed while checking it against live data: the merge workflow in the proposals repo currently snapshots the PR discussion as a single markdown file (`proposal_discussion.md`, next to `proposal.md` and `outcomes.json` in the same directory CID). On a real Q2 proposal CID, `/discussion.json` and `/summary.md` return 404, while `/proposal_discussion.md` returns the content. So with the current uploader the section would show the empty state for all existing proposals.
> 
> Two possible ways to close the gap: update the uploader to also emit `discussion.json`, or add a markdown fallback that renders `proposal_discussion.md` when `discussion.json` is absent. The fallback has the advantage that every proposal produced by the current uploader becomes displayable without touching the proposals repo. The existing Q2 ones already contain `proposal_discussion.md`, and the Q3 ones will too when their PRs merge.

Thank you @christian-rogobete for the detailed feedback.  I will look into this and fix all the reviews pointed out.

Oluwatos94:
> Thanks for checking this out @christian-rogobete! I agree that we should just use the current file naming. For summary as this is a new thing it's ok to use any name and yes that has to be fully optional.
> 
> @Oluwatos94 Please check locally with the existing project `stellarpga`.

Noted! 📝

dapp/src/service/ProposalService.ts
```diff
@@ -3,10 +3,23 @@ import {
   fetchTextFromIpfs,
   fetchJsonFromIpfs,
 } from "utils/ipfsFunctions";
-import type { Proposal, ProposalOutcome } from "types/proposal";
+import type { Proposal, ProposalOutcome, DiscussionPost } from "types/proposal";
 
 const PROPOSAL_MD_PATH = "/proposal.md";
 const OUTCOMES_JSON_PATH = "/outcomes.json";
+const DISCUSSION_JSON_PATH = "/discussion.json";
```
Shadow-MMN: `DISCUSSION_JSON_PATH` points to `/discussion.json`, but that file isn't produced by the proposals uploader. Existing proposals store the discussion as `/proposal_discussion.md`, so this path should be updated to match the current IPFS layout.

dapp/src/service/ProposalService.ts
```diff
@@ -18,18 +31,61 @@ async function fetchProposalFromIPFS(cid: string) {
...
+      typeof raw.timestamp === "number"
+        ? raw.timestamp
+        : String(raw.timestamp ?? ""),
+    body,
+    ...(typeof raw.source === "string" ? { source: raw.source } : {}),
+  };
+}
+
+/** Fetches the discussion thread (discussion.json) from IPFS. */
+export async function fetchProposalDiscussionFromIPFS(
```
Shadow-MMN, line 46: This implementation assumes the discussion is stored as JSON (`fetchJsonFromIpfs` + `normalizePost`), but the uploader actually stores it as a markdown file (`proposal_discussion.md`). This should mirror `fetchProposalFromIPFS` by fetching the markdown as text instead of parsing JSON.

dapp/src/components/page/proposal/DiscussionSection.tsx
```diff
@@ -0,0 +1,138 @@
+import Button from "components/utils/Button";
+import DOMPurify from "dompurify";
+import Markdown from "markdown-to-jsx";
+import React, { useState } from "react";
+import type { DiscussionPost } from "types/proposal";
+import { getIpfsBasicLink } from "utils/ipfsFunctions";
+
+interface DiscussionSectionProps {
+  discussion: DiscussionPost[] | null;
```
Shadow-MMN: The `discussion` prop is currently typed as `DiscussionPost[] | null`, but after reading `proposal_discussion.md` it will become a single markdown string. The component can render that string as one markdown block instead of iterating over individual posts.

website/docs/developers/governance.mdx
```diff
@@ -37,6 +37,30 @@ graph LR;
...
 
+### Proposal IPFS directory
+
+The proposal's `ipfs` field is the CID of a directory holding the proposal's off-chain content. The dapp reads the following files from it:
+
+| File | Required | Purpose |
+| --- | --- | --- |
+| `proposal.md` | yes | Proposal description (Markdown). |
+| `outcomes.json` | yes | Descriptions/XDR for the approved, rejected and cancelled outcomes. |
+| `discussion.json` | no | Discussion thread rendered on the proposal page. |
```
Shadow-MMN: The documentation still references `discussion.json`, but the proposals uploader stores the discussion as `proposal_discussion.md`. The docs should be updated to reflect the actual file layout used by existing proposals.

Shadow-MMN (changes requested):
After looking into this and checking existing proposal data, I think this needs a different approach.

The current implementation expects the discussion to come from `discussion.json`, but the proposals uploader actually snapshots the PR discussion as a markdown file named `proposal_discussion.md`, alongside `proposal.md` and `outcomes.json`.

I verified this against a real proposal CID (`QmdNHt6VvcMzMcmQjAKQS1cGwqsD47bctQy5T9VBs2oHuD`) from the `stellarpga` project on testnet. `/discussion.json` and `/summary.md` both return 404, while `/proposal_discussion.md` contains the discussion as expected.

Because of that, this implementation would always fall back to the empty state for existing proposals, even though the discussion is available.

As discussed with @tupui, I think we should read the thread from `proposal_discussion.md` (markdown) instead of `discussion.json`. The discussion summary can remain optional (the filename isn't important since it's new). That keeps this compatible with all existing proposals without requiring any changes to the proposals repository.

While making this change, there are a few related updates that should happen as well:

- Remove the `discussion.json`-specific pieces from `ProposalService` (`DISCUSSION_JSON_PATH` and `normalizePost`).
- Update `DiscussionSection` and `ProposalPage` so the discussion is represented as a single `string | null` instead of `DiscussionPost[]`.
- Remove the now-unused `DiscussionPost` type.
- Update the documentation to reference `proposal_discussion.md`.
- Adjust the tests to mock `fetchTextFromIpfs` and `/proposal_discussion.md` instead of the JSON endpoint.

#### Current changes look like this:

Image


#### What it would look like after changes on review: 
Image

Shadow-MMN:
And there are some merge conflicts, please do resolve them in the next commit.

tupui (approved):
Thanks a lot! Getting this in and will push some fixes for the bindings

